PR #18, opened 2016-03-16
Fix method doc loading

MugeSo:
Allow method doc as REDME style when method handler is Array as in README.

Following style is used in README.
```javascript
module.exports.post = [function (res, req, next) { next();}, post];
function post(res, req, next) { res.status(200}.json({message:'ok'});}
module.exports.post.apiDoc = {
  // api doc here
};
```

On the other hand, implementation and test uses this:
```javascript
module.exports.post = [function (res, req, next) { next();}, post];
function post(res, req, next) { res.status(200}.json({message:'ok'});}
post.apiDoc = {
  // api doc here
};
```

This PR allows both style. And for BC, latter style has priority.

### DIFF
--- a/index.js
+++ b/index.js
@@ -288,9 +288,9 @@ function getAdditionalMiddleware() {
 }
 
 function getMethodDoc(methodHandler) {
-  return Array.isArray(methodHandler) ?
-    methodHandler.slice(-1)[0].apiDoc || methodHandler.apiDoc :
-    methodHandler.apiDoc;
+  return methodHandler.apiDoc || (Array.isArray(methodHandler) ?
+    methodHandler.slice(-1)[0].apiDoc :
+    null);
 }
 
 function sortApiDocTags(apiDoc) {
--- a/index.js
+++ b/index.js
@@ -102,9 +102,7 @@ function initialize(args) {
     Object.keys(pathModule).filter(byMethods).forEach(function(methodName) {
       // methodHandler may be an array or a function.
       var methodHandler = pathModule[methodName];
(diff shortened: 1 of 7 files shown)


MugeSo:
@jsdevel If just readme is incorrect, reject this PR and fix it :wink:

jsdevel:
@MugeSo I believe this has been fixed in the latest version.  What version are you using locally?

jsdevel:
NVM.  I see it now.

index.js
```diff
@@ -289,6 +287,12 @@ function getAdditionalMiddleware() {
   }
 }
 
+function getMethodDoc(methodHandler) {
+  return Array.isArray(methodHandler) ?
+    methodHandler.slice(-1)[0].apiDoc || methodHandler.apiDoc :
```
jsdevel: How about:
```javascript
  return methodHandler.apiDoc || (Array.isArray(methodHandler) ?
     methodHandler.slice(-1)[0].apiDoc :
     null);
```

MugeSo: Yes, it's better if you don't mind BC.

The order becomes problem only when:
```js
module.exports.post = [function (res, req, next) { next();}, post];
function post(res, req, next) { res.status(200}.json({message:'ok'});}
post.apiDoc = {
  // api doc 1
};
module.exports.post.apiDoc = {
  // api doc 2
};
```

But, this case seems very rare. So, no need to mind BC.

jsdevel: I feel it is a user error if they define `apiDoc` in both places.

jsdevel:
Thanks @MugeSo !

PR #20, opened 2016-03-17
	partial fix for #17: Support external schema reference. 

### LINKED ISSUE #17: Support external schema reference
## Problem
Now my api receives json schema object.

so, i wrote:
```javascript
// in handler
post.apiDoc = {
    description: "create new schema definition",
    operationId: "createSchema",
    parameters: [
        {
            in: 'body',
            name: 'schema',
            schema: {$ref: '#/definitions/schema' }
        }
    ],
    responses: {
        "201": { $ref: '#/responses/created' },
        "default": { $ref: "#/responses/error" }
    }
}

// in apidoc.js#definitions
{ 
    "schema": { "$ref": "http://json-schema.org/schema" }
}  
```

Then i sent request to it and got:
```json
{
    "status":400,
    "errors":[
        {
            "message":"no such schema <http://json-schema.org/schema>",
            "schema":{"$ref":"http://json-schema.org/schema"},
            "location":"body"
        }
    ]
}
```

## Suggestion

1.  support external schema
2.  support maping external id to local file (for when can not connect to server hosting the external schema or the url don't serve schema json.)
3.  keep external reference as is in exposed api doc, not expanded loaded schema.

jsdevel:
* Supporting  in parameters and responses.
* Only supporting local references at this time to avoid async operations.
* Bumping coverage back up.
* Adding a warning message in the console for $ref: #/definitions/... in responses object.

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,6 +8,8 @@ var loggingKey = require('./package.json').name + ': ';
 var path = require('path');
 var buildValidationMiddleware = require('express-openapi-validation');
 var buildResponseValidationMiddleware = require('express-openapi-response-validation');
+var PARAMETER_REF_REGEX = /^#\/parameters\/(.+)$/;
+var RESPONSE_REF_REGEX = /^#\/(definitions|responses)\/(.+)$/;
 var validateSchema = require('openapi-schema-validation').validate;
 var METHOD_ALIASES = {
   get: 'get',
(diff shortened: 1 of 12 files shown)


jsdevel:
@MugeSo can you review this?  I know `definitions` probably shouldn't be available for `$ref` in responses, but it would break BC if we enforced only de-referencing `responses` at this time.

MugeSo:
It's good for local reference.
How about external reference?

jsdevel:
External reference will require more thought / work unfortunately.  The async nature of handling that would require a _lot_ of workarounds at this time.

MugeSo:
I think https://github.com/kogosoftwarellc/express-openapi/issues/17#issuecomment-197658250 works well.

MugeSo:
Maybe my comment confused you...
`json-ref-lite` is sync.

jsdevel:
OK.  If you can submit a PR for external refs then that would be great.  For now I'm going to merge this.

MugeSo:
I see, thanks :+1:

PR #21, opened 2016-03-18
Fix typo in comment

MugeSo:
Maybe, `gument` is inserted accidentally unless it is a new word such as even google doesn't know.


jsdevel:
Good catch!

PR #22, opened 2016-03-18
Support external response & parameter definitions

MugeSo:
refs #17 

This excepts external reference of schema.
Thus, #17 should be kept open.


MugeSo:
Oh, json-ref-lite is missing. Fix it soon.

jsdevel:
@MugeSo in thinking about this further, I'm not sure we should rely on `json-ref-lite`.  I'm almost thinking we should add to `args`.  Maybe `args.additionalSchemas`.  Those will need to be loaded into `express-openapi-validation` etc.

jsdevel:
this was handled in #30 .  Thanks again @MugeSo !

PR #28, opened 2016-03-28
Typing: Fix `Request` interface to follow coercion result.

kanreisa:
`request.get()`, `request.header()`, `request.headers[]` may not returns string only.

because:
https://github.com/kogosoftwarellc/express-openapi-coercion/blob/33a4bdbf95b26675e26db8c4b80ad5a3abc9a013/index.js#L41-L80


jsdevel:
Thanks @kanreisa !

PR #32, opened 2016-04-02
Support middleware builder

MugeSo:
Add middleware builder support.
This is useful to add middleware dynamically according to `methodDoc` or `apiDoc` for each operations.
In fact, I implemented this to add security middleware according to [`Security Requirement Object`](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#security-requirement-object). You can see simple example of this in sample application `with-middleware-builder` added in this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,8 @@ https://github.com/kogosoftwarellc/express-openapi/tree/master/test/sample-proje
   * See [Work with TypeScript](#work-with-typescript)
 * Supports external schema references
   * See [args.externalSchemas](#argsexternalschemas)
+* Supports middleware builders
+  * See [args.middlewareBuilder](#argsmiddlewarebuilder) and [Middleware Builder](#middleware-builder)
 
 ## Example
 
@@ -87,6 +89,8 @@ you can use these properties in said operation's apiDoc.  See full examples in t
 [./test/sample-projects/](
 https://github.com/kogosoftwarellc/express-openapi/tree/master/test/sample-projects)
 directory.
+And you can also add middleware using [Middleware Builder](#middlewarebuilder), which supplied as
+`middlewareBuilder` property of initialize argument or operations.
 
 ### Supported vendor extensions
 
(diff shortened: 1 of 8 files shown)


jsdevel:
This is actually a _really_ cool approach!  I hadn't thought of handling it like this.

Couple of questions:

* How does this affect additional middleware?  I kind of feel additional middleware specified via `x-express-openapi-additional-middleware` should be added to the middleware array _before_ the `middlewareBuilder`s have a chance to modify the middleware array.
* What do you think about defining `disableGlobalMiddlewareBuilder` in the apiDoc?  I like the terseness of `disableGlobalMiddlewareBuilder`.  What do you think about `x-express-openapi-disable-middleware-builder`?  This would be a scoped extension.
* Is `middlewareBuilder` the name we want/like?  I was thinking about some alternatives and would like your input:
  * `middlewareTransformer`
  * `middlewareHandler`
  * `onMiddlewareAdded`

MugeSo:
@jsdevel 
Maybe, some of your question is answered in README in 0874885

MugeSo:
> What do you think about x-express-openapi-disable-middleware-builder?

Truth to tell, I'm worried since before, I think [vendor extension](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#specification-extensions) is not for internal thing but for API uses.
So, sorry to say, I think other `x-express-opeapi-*` should be deprecated.

> Is middlewareBuilder the name we want/like?

I say "like negatively". In fact I bothered about this when I named it.
Any of name you suggested and also middlewareBuilder is not suit :worried:

jsdevel:
If your main use case right now is security, perhaps we should implement something like `args.securityMiddleware`?  I'm thinking something like this:


```javascript
var apiDoc = {
  swagger: '2.0',
  /* ... */
  security: {
    // api wide security, may be overridden in operations.
    api_key: []
  },
  securityDefinitions: {
    api_key: {
      type: 'apiKey',
      name: 'api_key',
      in: 'header'
    }
  }
};

var initializedApp = openapi.initialize({
  app: app,
  apiDoc: apiDoc,
  securityMiddleware: {
    api_key: function(req, res, next) {
      if (!req.headers.api_key) var err ={status: 401};
      next(err);
    }
  }
});
```

MugeSo:
@jsdevel not only for security.
For example, file upload using [multer](https://www.npmjs.com/package/multer).

I think this approach keeps `express-openapi` simple.

jsdevel:
File types are another case where the swagger spec isn't supported in `express-openapi`.  The [spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameter-object) has a `file` type as I'm sure you're aware.

The main reason why I like [this](https://github.com/kogosoftwarellc/express-openapi/pull/32#issuecomment-204652245) approach, is that I feel it's more declarative.  `middlewareBuilder` feels very imperative to me.

If security is the main use case for `middlewareBuilder`, then I'd rather use the `args.securityMiddleware` approach.

jsdevel:
Also, for security, the [spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#schema) appears to state that operation `security` should override api doc `security` with this statement:

```
A declaration of which security schemes are applied for the API as a whole. The list of values describes alternative security schemes that can be used (that is, there is a logical OR between the security requirements). Individual operations can override this definition.
```

I'll submit a PR shortly to implement `security`.

jsdevel:
One more point for `securityMIddleware`, now that Swagger UI fully supports security, we gain in functionality there by encouraging reliance on `securityMiddleware`.

MugeSo:
This also allows to add middleware before express-openapi middleware. Currently additional middleware is only after them.

It's nice for me, if you implement securityMiddleware.
However, even if so, I think middlewareBuilder is needed.

As you say this is imperative, but it is necessary to extends express-openapi.
Without this extensibility, we need implement all methoDoc related middleware in express-openapi.
And developer who use express-openapi need to wait it or implement it by them selves and send PR.
I feel it's nonsense.

MugeSo:
@jsdevel Don't forget scope handling when you implement securityMiddleware.

MugeSo:
I found more declarative solution to this.

apply following high-order function to all user defined handlers.
```
function getHandler(handler, methodDoc, apiDoc) {
   return handler.length === 2 ? handler(methodDoc, apiDoc) : handler;
}
```
This based on express's request handler and error handler has 3 and 4 arguments.

MugeSo:
> This also allows to add middleware before express-openapi middleware. Currently additional middleware is only after them.

This is needed to add operation related body parser.
Body parser should be called before validation.

jsdevel:
Could bodyParser be added for the app as a whole?  I see how `middlewareBuilder` would limit it in scope to just the API though.

MugeSo:
> Could bodyParser be added for the app as a whole?

Recall the fact that [OpenAPI spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#operation-object) allows `consumes` property of `Operation Object`.
This means we can limit the operation use specific body parser. e.g.  basic operations consumes JSON and a special operation consumes YAML.
On this case, is it correct waty to add  yaml bodyparser to the app?

jsdevel:
@MugeSo sorry there hasn't been any movement on this.  How are things going with `securityHandlers`?  Has that maybe eliminated the need for these changes?

MugeSo:
> Has that maybe eliminated the need for these changes?

No, this is still needed yet to resolve operation specified body parser (against `consume` property) and extensibility.

jsdevel:
What if we added support for consumes?  Seems like something we should add.

MugeSo:
This issue can be close when #39 is implemented, if you drop extensibility.

jsdevel:
@MugeSo I just published support for consumes (#39).  Can you check it out and let me know what you think?

Thanks for your patience!

MugeSo:
@jsdevel I think operation should be able to have consume middleware.

It seems [multer](https://github.com/expressjs/multer) prefers to set file field name to accept.
Of course I know express-openapi does not handle file upload yet.
However, multer also parses text fields in `multipart/form-data` into `req.body` and express-openapi can validate it.

MugeSo:
BTAIM, I think this PR can be closed by ec3b0c0.
Thanks!

jsdevel:
I'd really love to have file uploads working at some point.  Still not sure on the best approach.

Thanks for all your feedback!

PR #33, opened 2016-04-02
Adding args.securityHandlers.

jsdevel:
@MugeSo can you review this?  I'm wondering what the best way to handle scope is (I've never even used scopes before so any info you can give me would be great).

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,8 @@ and validation.
   * See [express-openapi-validation](https://github.com/kogosoftwarellc/express-openapi-validation)
 * Leverages openapi response definitions to provide `res.validateResponse` tailored to a particular route.
   * See [express-openapi-response-validation](https://github.com/kogosoftwarellc/express-openapi-response-validation)
+* Leverages security definitions for security management.
+  * See [express-openapi-security](https://github.com/kogosoftwarellc/express-openapi-security)
 * Validates api documents.
   * See [openapi-schema-validation](https://github.com/kogosoftwarellc/openapi-schema-validation)
 * Configurable Middleware.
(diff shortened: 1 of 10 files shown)


MugeSo:
It seem not good.

At first,
`security` properties is Array of [`Security Requirement Object`](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#securityRequirementObject) and it means authentication is pass when one of element is success.
And `Security Requirement Object` has one or more auth and scopes. Authentication will pass when all of element in it.

However this implementation flatten them.  It's wrong.

MugeSo:
>  I'm wondering what the best way to handle scope is

I recommend to use higher-order function returns middleware.

```javascript
openapi.initialize({
   apiDoc: apiDoc,
   app: app,
   securityMiddleware: {
     oauthSchema: function(scopes) {
       return function(req, res, next) {
         /* do something */
         next();
       }
     }
   }
});
```

This is needed only for schema whom type is OAuth.

And if you want to allow both of high-order function and handler function,
you can distinguish them using `handler.length` is 1 or 3.
```javascript
handler = handler.length === 1 ? handler(scopes) : handler;
```

jsdevel:
@MugeSo what if we wrapped the middleware and added `securityScopes` to the `req` object?  Then the middleware would have what it needs.

jsdevel:
> this implementation flatten them. It's wrong

I agree.  I believe each object in the array represents a set of security requirements that must all pass.  If any of them do not, then the schemes in the next object are evaluated.  I'm not sure what the best way to handle that would be though.

MugeSo:
> what if we wrapped the middleware and added securityScopes to the req object?

Imagine using existing authentication library. It handles scopes added to the request? ― No.
As a result, we need convert that into library could handle.
Furthermore, conversion must be executed every request because scopes is in request.

On the other hand, the solution using higher-order function, conversion can be one.

MugeSo:
> I'm not sure what the best way to handle that would be though.

This is why I [said](https://github.com/kogosoftwarellc/express-openapi/issues/3#issuecomment-180188366) security is too complex to handle in `express-openapi`

jsdevel:
@MugeSo I've moved this to [express-openapi-security](https://github.com/kogosoftwarellc/express-openapi-security).  I believe it handles security correctly, or at least in adherence with openapi docs.

MugeSo:
@jsdevel It seems work well with openapi docs :smile:
Please remember that this implementation may not suit well a existing authentication library.

MugeSo:
@jsdevel I wonder whether or not 401 is correct as a status code when scope does not match. It should be 403, shouldn't it?

And I found we need send [`WWW-Authenticate'](http://tools.ietf.org/html/rfc7235#section-4.1) when send status 401.
See http://tools.ietf.org/html/rfc7235#section-3.1

MugeSo:
`WWW-Authenticate` header could generate from `securityDefinition`

MugeSo:
Registered schema list is [here](http://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml)

MugeSo:
As was expected we should send 403 when get scope mismatch. 
See http://tools.ietf.org/html/rfc6750#section-3.1

MugeSo:
> Registered schema list is here

As you know, [openapi allows](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields-15) only basic, api-key and oauth2.

MugeSo:
OAuth2 also support [Mac schema](https://tools.ietf.org/html/draft-ietf-oauth-v2-http-mac-01)

jsdevel:
@MugeSo so I think we can handle the various auth schemes if the first arg is an error object, similar to the following:

```
{
  status: 401,
  headers: {
    'www-authenticate': 'Basic'
  }
}
```

Thoughts?

MugeSo:
@jsdevel You mean errors are merged, right?

It's summarized as follows:

Errors in `Security Requirement Object`:
- success when all of elements success
- any of error has status 401 then status is 401
- `www-authenticate` are merged
- response body?

Errors in `security` property (Array of `Security Requirement Object`):
- success when any of elements success
- any of error has status  403 then status is 403
- `www-authenticate` are merged
- response body?

Issue #40, opened 2016-05-11
Account for deeply nested apps in basePath.


willist:
When express-openapi is mounted on a sub app, the basePath ignores the mount path of the parent app.
```javascript
var app = express(),
    sub = express(),
    sub2 = express();

openapi.initialize({
  apiDoc: require('./api-doc.js'),
  app: sub2,
  routes: './api-routes'
});

sub.use('/subapp2', sub2);
app.use('/subapp', sub);
```

If ``./api-doc.js`` has a basePath of ``/v1``, I would expect the basePath in the generated api doc in this case to be ``/subapp/subapp2/v1``. Instead, the basePath of the api doc is ``/subapp2/v1``, ignoring the highest mount path.

jsdevel:
@willist can you submit a PR for this?

willist:
Fixed by #41

jsdevel:
Thanks for closing!

PR #41, opened 2016-05-11
Update basePath at request time.

willist:
Handle nested apps (see issue #40) by updating the basePath when the apiDoc is requested.

### DIFF
--- a/index.js
+++ b/index.js
@@ -212,6 +212,7 @@ function initialize(args) {
   if (exposeApiDocs) {
     // Swagger UI support
     app.get(basePath + docsPath, function(req, res) {
+      apiDoc.basePath = req.baseUrl + basePath;
       res.status(200).json(apiDoc);
     });
   }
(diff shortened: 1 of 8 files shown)


jsdevel:
Looks like tests are failing.  Can you take a look?

willist:
It looks like I missed some unstaged changes in the first commit.  Tests are now passing on 4.x and 5.x.  I'll need to dig into the 0.10 and 0.12 failures though.

jsdevel:
also, I'm surprised that this isn't working already.  See https://github.com/kogosoftwarellc/express-openapi/blob/master/index.js#L223

willist:
Okay, it looks like there were two separate issues in my project.

1. The parent app was an instance of Router instead of Application.  Router does not appear to have a mountpath, which is why I wasn't seeing any modification to the basePath.
2. I have several nested applications above express-openapi.  I am serving the api doc at ``/api/nested/v1/api-docs``, where each level of that path is a separate application.  Even when I resolve the first issue by switching from Router to Application, my basePath is ``/nested/v1`` instead of ``/api/nested/v1`` due to limitations of mountpath.

This PR resolves both issues, but is missing unit tests that accurately reflect those issues.  I'll add those in the morning and update this PR.  I'll also look into and fix issues with 0.10 and 0.12.

I've also updated #40 to better explain item 2 above.

jsdevel:
> This PR resolves both issues, but is missing unit tests that accurately reflect those issues. I'll add those in the morning and update this PR. I'll also look into and fix issues with 0.10 and 0.12.

Thanks!  Looking forward to merging this.

willist:
I added test cases and fixed the failures on 0.10 and 0.12.

PR #46, opened 2016-09-02
Supporting args.routes. closes #45

### LINKED ISSUE #45: modular approach
Hello, thank you for the excellent framework. I have a question about using it to construct modular express applications, where multiple routes are defined in individual npm modules, and the final product is composed from these. Can I call openapi.initialize() multiple times in a single app, once for each module? This implies that instead of a single routes folder, I may have several...

jsdevel:


troymolsberry:
Initial User Acceptance testing passed. THANK YOU for such a quick turnaround on this!

jsdevel:
> THANK YOU for such a quick turnaround on this!

Any time!  Thanks for contributing!

PR #50, opened 2016-09-08
adding support for dependency injection into routes

troymolsberry:
This change fundamentally modifies whether or not route modules return an object (current behavior) or return a function that can accept arguments when routes are loaded. A new, optional args attribute named routesDependencyInjection controls whether or not routes are loaded as objects or as functions...


jsdevel:
I like where this is going!  What is the main use case?

troymolsberry:
since express-openapi uses a folder tree for each route, this provides a
way to inject things like dataproviders or service providers into each
route rather than have them each require what they need individually.

jsdevel:
Ah.  I like it!

Couple of suggestions:

* Instead of args.routesDependencyInjection as an array, how about args.dependencies as an object.  Then we can read the argument names of the function by parsing the result of toString.

jsdevel:
I also feel this would make testing easier for those who want to unit test the handlers instead of using supertest.

jsdevel:
Also, can you update the section about routes to also include an example with injection?

troymolsberry:
sure thing. I also hate the name args.routesDependencyInjection...should it
be ?

args.dependencies = {
     "dataprovider": { ... },
     "config": { ... }
};


The weird thing about an object versus an array is that as far as I know
you cannot guarantee ordering of properties in objects, but a user of this
capability may have signatures they have to match, which using an array
gives you that capability. Thoughts on that?

Also, I'd be happy to accept an array OR an object.

troymolsberry:
also, since we are thinking this through, let's be sure this is even
needed. Alternatively to doing dependency injection, couldn't I effectively
just attach my "dependencies" to the express object and pull them from the
req parameter in each route handler? Sorry for the second guess here, just
making sure this s a useful change.

troymolsberry:
on further thought, I believe there is a use case for constructor injection
of dependencies into routes modules, versus attaching things to the express
object. The attaching method only provides access to the dependencies from
within routes, which is different than having everything inside a closure.

jsdevel:
@troymolsberry the beauty about an object, is that the ordering really doesn't matter.  Check this out:

```
function myHandler(myService, someOtherService) {
  return myService.hello() + ' ' + someOtherService.world();
}

var DEPENDENCY_REGEX = /^function\s[^(]*\(\s*((?:(?!\))[\s\S])*)\)/m;

var dependencies = {
  myService: {
    hello: function() {return 'hello';}
  },
  someOtherService: {
    world: function() {return 'world';}
  }
};
var declaredDependencies = DEPENDENCY_REGEX.exec(myHandler.toString());
var injectables = [];
if (declaredDependencies) {
  injectables = declaredDependencies[1].trim().split(/\s*,\s*/).map(function(name) {
    if (name in dependencies) {
      return dependencies[name];
    }
    throw new Error('No dependency found for ' + name);
  });
}
myHandler.apply(null, injectables);
```

README.md
```diff
@@ -286,6 +286,39 @@ function, or an array of business specific middleware + a method handler functio
...
+
+Example using named (object) dependency injection
+```javascript
+openapi.initialize({
+    apiDoc: require('./api-doc.js'),
+    app: app,
+    routes: [
+        path.resolve(__dirname, 'api-routes'),
+    ],
+    dependencies: {"named parameter 1": {/* dependency*/}, "named parameter 2": {/* dependency*/}}
```
jsdevel: can you use valid argument names here, and show how it would be mapped in a handler file?

README.md
```diff
@@ -286,6 +286,39 @@ function, or an array of business specific middleware + a method handler functio
...
+
+Example using positional (array) dependency injection
+```javascript
+openapi.initialize({
+    apiDoc: require('./api-doc.js'),
+    app: app,
+    routes: [
+        path.resolve(__dirname, 'api-routes'),
+    ],
+    dependencies: [{ /* dependency passed as positional argument 1 */}, {/* dependency passed as positional argument 2 */}]
```
jsdevel: I'm not sure I like the array approach.  Feels like it couples the dependencies listed here with the ordering of the arguments declared in the handler files.

troymolsberry:
removed the array form of injection and enhanced the tests and documentation. starting down the path of injecting a dataprovider into some routes.

jsdevel:
Thanks @troymolsberry!  This was a fantastic idea!

jsdevel:
Published as `v0.27.0`.

troymolsberry:
Thank you for the excellent framework!

PR #57, opened 2016-10-17
Static routes before dynamic routes

casras:
I have following routes:

pathx/{id}
pathx/latest

And it seems that the dynamic pathx/{id} currently takes precedence, over the static pathx/latest, which might not be desired?


jsdevel:
Thanks @casras !

Issue #12, opened 2016-03-03
Alterative route declaration


DaleJefferson:
The only thing I don't like about express-openapi is the "convention based programming", I found very confusing and limiting.

I would prefer a simple javascript api doc with embedded handlers.

```
openapi.initialize({
    app: app,
    swagger: '2.0',
    basePath: '/v1',
    info: {
        title: 'App',
        version: '1.0.0',
    },
    paths: {
        '/users': {
            get: {
                parameters: [],
                responses: {
                    200: {
                        description: 'Get all users',
                        schema: {
                            type: 'array',
                            items: {
                                title: 'User',
                                type: 'object',
                                properties: {
                                    name: {
                                        type: 'string'
                                    }
                                }
                            },
                        }
                    }
                },
                handler(req, res) {
                    res.json([{name: 'Dale'}]);
                }
            },
            post: {
                handler(req, res) {
                    res.status(201);
                }
            }
        },
        '/users/{userId}': {
            get: {
                handler(req, res) {
                    res.json({name: 'Dale'});
                }
            }
        }
    }
});
```

You can use modules to split up the code, you have the full javascript language to generate routes or do whatever you like.

```
openapi.initialize({
    app: app,
    swagger: '2.0',
    basePath: '/v1',
    info: {
        title: 'App',
        version: '1.0.0',
    },
    paths: {
        '/users': require('./routes/users'),
        '/messages': {
            'get': require('./routes/messages').get
        },
        '/books': createCrudEndpoints("Books"),
        ...require('./lotsOfEndpointsUsingSpread')
    }
});
```

The existing convention based system could be used to build this structure internally.

jsdevel:
So you can define paths in `apiDoc.paths`, and `express-openapi` will inherit from it, except for handlers.

What you would consider limiting, I would consider freeing.

Consider that you never have to duplicate a path, as it's entirely inferred from fs.

jsdevel:
Also, you're documentation is kept up to date for users.  So if you delete a handler, you don't have to worry about your docs getting out of sync.  This is what bothers me about other frameworks that expect a monolithic json document or yaml doc to descript their apis.

jsdevel:
Re-opening as express-openapi should aim to be as un-opinionated as possible to follow express's purpose `Fast, unopinionated, minimalist web framework for Node.js`.

MugeSo:
[...]

PR #60, opened 2016-10-18
Add option to specify routes by hand

mika-fischer:
Format is { route: '/foo/{id}', handlers: require('./foo') }


README.md
```diff
@@ -671,6 +671,9 @@ post.apiDoc = {
 
 ```
 
+Alternatively, args.paths may contain route specifications of the form
```
jsdevel: I realize we'll end up internally merging these with `fs-routes`; however, on second look we should probably stay more inline with openapi concepts.  How about something like this? `{path: '/foo/{id}', module: require('./handlers/foo')}`

mika-fischer: OK, fine with me, I'll update the pull request accordingly.

mika-fischer: Done, have a look

jsdevel: Looks great!  Thank you so much for adding this!  I feel it gets us closer to resolving #12.

jsdevel:
Published as `v0.31.0` @mika-fischer

PR #67, opened 2017-01-26
Allow multiple named parameter in single path part

MugeSo:
This PR allows `/{foo}-{bar}/{file}.{ext}`

### DIFF
--- a/index.js
+++ b/index.js
@@ -399,7 +399,9 @@ function byProperty(property, value) {
 function byRoute(a, b) {
   if(isDynamicRoute(a.path) && !isDynamicRoute(b.path)) return 1;
   if(!isDynamicRoute(a.path) && isDynamicRoute(b.path)) return -1;
-  return a.path.localeCompare(b.path);
+
+  // invert compare to keep that /{foo} does not beat /{foo}.{bar}
+  return -1*a.path.localeCompare(b.path);
 }
 
 function isDynamicRoute(route) {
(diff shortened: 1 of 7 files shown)


index.js
```diff
@@ -399,7 +399,9 @@ function byProperty(property, value) {
 function byRoute(a, b) {
   if(isDynamicRoute(a.path) && !isDynamicRoute(b.path)) return 1;
   if(!isDynamicRoute(a.path) && isDynamicRoute(b.path)) return -1;
-  return a.path.localeCompare(b.path);
+
+  // invert compare to keep that /{foo} does not beat /{foo}.{bar}
+  return -1*a.path.localeCompare(b.path);
```
jsdevel: space around operators e.g. `-1 * a.path......`

MugeSo:
> Can you also add this to the README? 

I don't know which section should I write it in.

jsdevel:
> I don't know which section should I write it in.

How about in a new description under [this heading](https://github.com/kogosoftwarellc/express-openapi#operation-parameters).  There's currently just a bullet point underneath that.  We should probably have a sub heading titled `Path Configuration` above the existing bullet point.

jsdevel:
On second thought, how about a new `Path Parameters` section above `Operation Parameters` that explains how `{param}` works in the filename as well as an example of multiple params in a path part.

jsdevel:
@MugeSo any updates?

jsdevel:
@MugeSo if you can add documentation later I'd appreciate it.  Thanks for the change!

Issue #79, opened 2017-12-15
Using "$ref: '#/definitions/...'" for responses has been deprecated.


XuluWarrior:
All the examples in the test suite have API definitions similar to 

`
  responses: {
    default: {
      $ref: '#/definitions/Error'
    }
  }`

These are flagged as errors in the Swagger Editor.

A valid alternative is to use

`  responses: {
    default: {
      description: "Unexpected error",
      schema: {
        $ref: '#/definitions/Error'
      }
    }
  }
`

jsdevel:
closed by #80

PR #80, opened 2017-12-15
#79 Using "$ref: '#/definitions/...'" for responses has been deprecated.

XuluWarrior:


jsdevel:
Thanks @XuluWarrior !

PR #81, opened 2017-12-28
Ignore test/spec files when loading paths

marcos-diaz:
Is a common practice in JS projects to place test files at the same folder than the files with the actual implementation, but with `*.test.js` or `*.spec.js` suffix.

This PR make the library to ignore these files when using `initialize()`, otherwise if there is any test file in the paths directory it will fail since lacking usual test-framework functions (ie: `describe`, `it`).

Since I see no possible scenario in which someone needs to opt-in to this behaviour, I considered there is no point to add additional args.


marcos-diaz:
Refactored as requested and added tests.

test/sample-projects/with-spec-in-routes/app.js
```diff
@@ -0,0 +1,31 @@
...
+var path = require('path');
+var cors = require('cors');
+
+app.use(cors());
+app.use(bodyParser.json());
+
+openapi.initialize({
+  apiDoc: require('./api-doc.js'),
+  app: app,
+  paths: path.resolve(__dirname, 'api-routes'),
```
jsdevel: can we add an `ignoreRegex` property to the `initialize` args?

marcos-diaz:
Added `pathsIgnore` args.
Please review again.

marcos-diaz:
Would be nice to have this released in NPM if there are no other changes in the pipeline.

Issue #82, opened 2017-12-28
Set apiDoc.host to running host


XuluWarrior:
When the api docs json is requested the basePath is automatically updated to include the baseUrl of the request.  It would make sense if the host is also updated to reflect the host in the request object.

This would allow the same code to run in different environments (dev, production, etc) without having to manually change the host.

jsdevel:
closed by #83

PR #83, opened 2017-12-28
#82 set apidoc host

XuluWarrior:


XuluWarrior:
Adding host to the api-docs breaks some tests as the deep compare with the test/fixtures fails.
I'm not sure the best way around this.  
Maybe some flag to enable/disable updating the host?
Or, perhaps only updating the host if the api-docs already has a value for host?

jsdevel:
Would it be possible to update the tests?  I like the idea of verifying the returned host header.

XuluWarrior:
The problem is that wrapping an express app in supertest results in it creating an "ephemeral" port for each test.  Which would mean that the port would change on each run.

I've taken a look and I see two ways around this.
The first is to call app.listen with a fixed port and wrap the listener instead e.g.
```
before(function() {
  app = require('./app.js').listen("54321");
});
```
This is the simplest change but I don't know if we need to be concerned with the possibility that the port might be in use.

The other option is to get the server address from the generated get request and modify the expected json with the host before comparing. e.g.
```
it('should expose <apiDoc>.basePath/api-docs', function(done) {
  var get = request(app)
    .get('/v3/api-docs');
  var host = get.url.match(/\/\/(.+:\d+)\//)[1];
  expectedApiDoc.host = host;
  get
    .expect(200, expectedApiDoc, done);
});
```

jsdevel:
You should be able to set the `Host` header right?

XuluWarrior:
I'm not sure what you mean.  Set Host on which header?

XuluWarrior:
I'm not sure what you mean.  Set Host on which header?

jsdevel:
```
  var get = request(app)
    .set('Host', 'foo.com')
    .get('/v3/api-docs');
  var host = get.url.match(/\/\/(.+:\d+)\//)[1];
  expectedApiDoc.host = host;
```

XuluWarrior:
That works.
I've set Host to "test-host" in all the tests that request api-docs and compare the full result.

jsdevel:
Thanks @XuluWarrior !  This is great!

PR #88, opened 2018-01-17
Lookup operation spec from main apiDoc when none is specified on the …

### LINKED ISSUE #87: Possibility to separate openapi-spec from http-controller
As far as i can see, the execution of any middleware like parameter validation happens based on the `apiDoc` property of a single http endpoint / route.

```js
module.exports.GET = () => {} // handler
module.exports.GET.apiDoc = {} // open api route schema
```

It would be nice if express-openapi could fallback to path-definitions passed to the `initialize({ apiDoc })` routine.

That way the developer could choose between having the route-handler and the api-spec in one file or separated.

I'd like to run the following setup:

```
api.js // openapi.initialize({ apiDoc: loadOpenApiSpec({ path: './' }), paths: './'})
api-doc.yml // swagger: 2.0 ...
paths/some-route/resource.js // module.export.GET = (req, res) => { ... }
paths/some-route/resource.yml // paths: { /some-route/resource: { get: { ... } } }
```
I would like to load and assemble the whole open-api spec with all path definitions just before passing it to `initialize`.

```js
const { sync: globSync } = require('glob');
const deepAssign = require('deep-assign');
const fs = require('fs');
const yaml = require('js-yaml');

module.exports = function loadOpenApiSpec({ version, path, specGlob = '/**/*.yml' }) {
  const files = globSync(specGlob, {root: path});

  const spec = {};

  files.forEach((file) => {
    deepAssign(spec, yaml.safeLoad(fs.readFileSync(file, 'utf8')));
  });

  spec.info.version = version;

  return spec;
};
```

So my idea is to keep backwards compatibility and to allow for populating the "global" spec down to a specific route. 

@jsdevel I could submit a PR if you like the idea.

alrik:
Provide an alternative and flexible way to pass in operation-level api spec.
If the handler does not provide a dedicated `apiDoc`, fallback to the operation-spec defined in the main apiDoc. 

* [x] Add support for passing path/operation-spec via `initialize({ apiDoc })` 
* [x] Adjust readme
* [x] Add test

resolves #87

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ This getting started guide will use the most fundamental concepts of OpenAPI and
     export default apiDoc;
     ```
 
-    You may be wondering why `paths` is an empty object literal.  We'll get to that in a second.
+    You may be wondering why `paths` can be an empty object literal.  We'll get to that in a second.
 
     This is all that is required for our API's main apiDoc.  To see the full list of values
     and options for the main apiDoc you can view [The Schema](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#schema).
@@ -154,8 +154,8 @@ This getting started guide will use the most fundamental concepts of OpenAPI and
 
 1. Create path handlers.
 
-    Our `paths` object was empty in the main apiDoc because `express-openapi` generates
-    it for us based on the location of our path handlers.  For this example we'll place
+    Our `paths` object was empty in the main apiDoc because `express-openapi` allows to
+    generate it for us based on the location of our path handlers.  For this example we'll place
     our path handlers under `api-v1/paths/`.
 
     Let's create a `worlds` path:
@@ -349,8 +349,9 @@ api.  An initialized api contains the following properties:
 |----|--------|-----------|
 |Object or String|Y|This is an OpenAPI (swagger 2.0) compliant document.  See the [OpenAPI-Specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md) for more details.|
 
-`args.apiDoc.paths` should be an empty object.  `express-openapi` will populate this
-for you.  This prevents you from defining your paths in 2 places.
+`args.apiDoc.paths` can be an empty object. In that case `express-openapi` will populate this
+for you based on your operation level apiDocs.
+But it is also possible to have just [one central apiDoc](test/sample-projects/basic-usage-with-central-apiDoc).
 
 `args.apiDoc.basePath` will add a prefix to all paths added by `express-openapi`.
 
--- a/README.md
+++ b/README.md
@@ -154,9 +154,9 @@ This getting started guide will use the most fundamental concepts of OpenAPI and
 
 1. Create path handlers.
 
-    Our `paths` object was empty in the main apiDoc because `express-openapi` allows to
-    generate it for us based on the location of our path handlers.  For this example we'll place
-    our path handlers under `api-v1/paths/`.
+    Our `paths` object was empty in the main apiDoc because `express-openapi` will generate it for us
+    based on the location of our path handlers.
+    For this example we'll place our path handlers under `api-v1/paths/`.
 
     Let's create a `worlds` path:
 
@@ -351,7 +351,7 @@ api.  An initialized api contains the following properties:
 
 `args.apiDoc.paths` can be an empty object. In that case `express-openapi` will populate this
 for you based on your operation level apiDocs.
-But it is also possible to have just [one central apiDoc](test/sample-projects/basic-usage-with-central-apiDoc).
+It is also possible to have just [one central apiDoc](test/sample-projects/basic-usage-with-central-apiDoc).
 
 `args.apiDoc.basePath` will add a prefix to all paths added by `express-openapi`.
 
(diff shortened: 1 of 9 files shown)


jsdevel:
@alrik any progress on this?

alrik:
@jsdevel Thanks for pinging me. I'm on it.

alrik:
@jsdevel this PR is ready for review.

alrik:
Ahoi @jsdevel,

did you already have have had time to look at this?
If have any changerequest, just tell me :)

I'm looking forward to get this out,
Cheers, Alrik

README.md
```diff
@@ -154,8 +154,8 @@ This getting started guide will use the most fundamental concepts of OpenAPI and
 
 1. Create path handlers.
 
-    Our `paths` object was empty in the main apiDoc because `express-openapi` generates
-    it for us based on the location of our path handlers.  For this example we'll place
+    Our `paths` object was empty in the main apiDoc because `express-openapi` allows to
```
jsdevel: Can you change this to:

```
Our `paths` object was empty in the main apiDoc because `express-openapi` will generate it for us based on the location of our path handlers.
```

README.md
```diff
@@ -349,8 +349,9 @@ api.  An initialized api contains the following properties:
 |----|--------|-----------|
 |Object or String|Y|This is an OpenAPI (swagger 2.0) compliant document.  See the [OpenAPI-Specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md) for more details.|
 
-`args.apiDoc.paths` should be an empty object.  `express-openapi` will populate this
-for you.  This prevents you from defining your paths in 2 places.
+`args.apiDoc.paths` can be an empty object. In that case `express-openapi` will populate this
+for you based on your operation level apiDocs.
+But it is also possible to have just [one central apiDoc](test/sample-projects/basic-usage-with-central-apiDoc).
```
jsdevel: Change this to:

```
It is also...
```

index.js
```diff
@@ -177,16 +177,20 @@ function initialize(args) {
...
     var originalPathItem = originalApiDoc.paths[openapiPath] || {};
     var pathItem = apiDoc.paths[openapiPath] || {};
-    var pathParameters = Array.isArray(pathModule.parameters) ?
-        [].concat(pathModule.parameters) :
-        [];
+    var pathParameters = pathItem.parameters || [];
+
+    // push all parameters defined in the path module to the path parameter list
+    if (Array.isArray(pathModule.parameters)) {
+      Array.prototype.push.apply(pathParameters, pathModule.parameters);
```
jsdevel: Change this to:

```
[].push.apply...
```

index.js
```diff
@@ -177,16 +177,20 @@ function initialize(args) {
...
+    }
+
     pathItem.parameters = pathParameters;
     apiDoc.paths[openapiPath] = pathItem;
 
     Object.keys(pathModule).filter(byMethods).forEach(function(methodName) {
       // operationHandler may be an array or a function.
       var operationHandler = pathModule[methodName];
-      var operationDoc = handleYaml(getMethodDoc(operationHandler));
+      var operationDoc = handleYaml(getMethodDoc(operationHandler)) || pathItem[METHOD_ALIASES[methodName]];
```
jsdevel: Shouldn't we allow YAML for both?

jsdevel:
@alrik sorry for taking so long to get back to you.  I feel this one is _almost_ there.

What about supporting operation doc files next to the handler files?  i.e.

```
/api/v3/users.js
/api/v3/users.docs.json
```

alrik:
Hey @jsdevel I hope to get your change requests done by tomorrow.
Regarding dedicated operation-doc files - sounds like a neat feature, but maybe out of scope of this PR :)

index.js
alrik: Just preference or any particular reason?

index.js
alrik: I'll check this, if we need to take care for yaml on `pathItem` at this point, or if it has already been handled.

index.js
jsdevel: Existing style + terseness + Array.prototype can be modified.

alrik: I'll make it consistent with the rest of the codebase.

```js
Array.prototype.push = 1;
[].push === 1 // true
```

jsdevel: Sadly, I remember the days of trying to decide between jQuery and http://api.prototypejs.org/language/Array/. :slightly_frowning_face:

index.js
alrik: As far as i see: pathItem is already an object. potential yaml has already been handled.

```js
var originalApiDoc = handleYaml(args.apiDoc);  // #L71 
var apiDoc = copy(originalApiDoc);                     // #L73
var pathItem = apiDoc.paths[openapiPath] || {};  // #179
```
see https://github.com/alrik/express-openapi/blob/feature/api-doc-fallback/index.js#L71 and https://github.com/alrik/express-openapi/blob/feature/api-doc-fallback/index.js#L73 and https://github.com/alrik/express-openapi/blob/feature/api-doc-fallback/index.js#L179

alrik:
@jsdevel done.

jsdevel:
Thanks @alrik !

PR #110, opened 2018-06-06
fixed link to sample projects in readme

magikMaker:
This fixes the link to the sample projects in the readme file so it also works on npmjs.org


jsdevel:
Thanks!  I'll publish in a bit

magikMaker:
👍great!

PR #127, opened 2018-06-28
Attempt to implement the 'nullable' keyword

lzear:
The [Swagger specification](https://swagger.io/docs/specification/data-models/data-types/) defines a syntax for nullable parameters that differs from JSON-Schema's
```yaml
# In Swagger / OAI 3: 
property: 
  type: string
  nullable: true

# In JSON-schema: 
property: 
  anyOf: 
    - type: string
    - type: 'null'
```
So far, "nullable" is not handled by the packages of this project.
Here is an attempt to convert the OAI syntax to the JSON-schema syntax for nullable parameters.

This should only be applied when using OpenAPI v3


lzear:
@jsdevel thanks a lot!
When is coercion needed exactly? The only use case that I can think of are path parameters.
I'm not sure I'll be able to update it, though it looks like an update is needed.

Also we can add some more tests to the request and response validators when the changes are published, for example by reverting https://github.com/kogosoftwarellc/open-api/commit/b5440025717baaba69937c5ddc0dfcefe77eab5f

jsdevel:
@lzear published `openapi-jsonschema-parameters@0.4.0`.

Coercion is needed to coerce incoming request parameters to their defined type i.e. `?foo=5` should be coerced to `{foo: 5}` instead of `{foo: '5'}`

PR #147, opened 2018-08-01
[openapi-types]: Fill out the OpenAPI v3 Typescript types

fluggo:
Now that the openapi attribute distinguishes v2 from v3, here are the types that belong to it.


jsdevel:
Brilliant!  Thanks @fluggo !

jsdevel:
published as `openapi-types@1.0.4`

PR #157, opened 2018-08-06
export SchemaObject "type" property type

bcrastnopol:
My typescript project wouldn't compile without casting my schema types as "string" | "number" | "boolean" | "object" | "integer" | "null". I created a type to match the one correct string literal locally but this way we can export the correct types in case the spec changes.


jsdevel:
published as `openapi-types@1.0.5`

PR #161, opened 2018-08-09
openapi-request-validator: use ajv instead of jsonschema for validation

srijs:
We have started to use this library for some of our production services, and during profiling we noticed that a large amount of the processing time for requests was being spent validating incoming requests.

To improve this situation, I have ported `openapi-request-validator` to use [`ajv`](https://ajv.js.org/) rather than `jsonschema`. In many cases this change results in speed-ups of 1-3 orders of magnitude, especially with more complex schemas involved.

I have run all positive test cases as benchmarks using the old and the new code, and published a gist with the raw results here: https://gist.github.com/srijs/b6d6e01931d6fe20d65c90f6d275f35c

The main reason `ajv` achieves these speed-ups is that it compiles the schema eagerly into a validation function, rather than interpreting the schema repeatedly on each validation.

I'm opening this PR to gauge your interest in upstreaming this change. There are some potential follow-up changes (e.g. porting `openapi-response-validator` and `openapi-schema-validator`), which I'm happy to include as part of this PR, or do as separate PRs.

Looking forward to discussing this!

### DIFF
--- a/packages/openapi-request-validator/README.md
+++ b/packages/openapi-request-validator/README.md
@@ -76,15 +76,15 @@ A function that transforms errors.
 This function is passed 2 arguments.
 
 ```
-  errorTransformer: function(openapiError, jsonschemaError) {
+  errorTransformer: function(openapiError, ajvError) {
     return {
       message: openapiError.message
     };
   }
 ```
 
-See the error format in [jsonschema](https://www.npmjs.com/package/jsonschema) for
-`jsonschemaError`.  `openapiError`s have the following properties:
+See the error format in [ajv](https://www.npmjs.com/package/ajv#validation-errors) for
+`ajvError`.  `openapiError`s have the following properties:
 
 * `errorCode` - A jsonschema error suffixed with `.openapi.validation`.
 * `location` - One of `body`, `headers`, `path`, or `query`.  Signifies where validation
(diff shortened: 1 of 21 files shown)


jsdevel:
@srijs this is simply fantastic!  Yes. Please follow up with a PR per package if you have time.  I'm going to publish this immediately.  Thanks again!

jsdevel:
published as `openapi-request-validator@2.0.0`.  Also, being that the errorTransformer is called with an ajvError now instead of a jsonschemaError, will this be a breaking change?

srijs:
Yeah, it's a breaking change due to the errors being mapped through.

I had some code to try and convert ajv errors into jsonschema errors, but that ended up pretty convoluted and felt like it wasn't worth the effort.

Another option could be to just not include the jsonschema or ajv errors when calling the error transformer, which would make the API more future proof, but I assume there was a good reason for including the jsonschema error in the first place?

jsdevel:
Yea, the original need was to make the lib more generic.  Ok.  I'll probably hold off on bumping versions in openapi-framework until we can bump all 3.  I'm guessing the only others that need to be converted are the schema validation package and the response validation package?

srijs:
Makes sense to me! I'm working on those two packages today, and should have PRs ready in a couple of hours if everything goes well. Once that's done, we'll need to cascade the version bumps through openapi-framework and express-openapi.

jsdevel:
Awesome!  Separate PRs for the remaining 2 packages if you please.  I'll handle bumping, publishing, and cascading.  Really can't thank you enough!  I'm excited for this change!

PR #164, opened 2018-08-10
openapi-response-validator: use ajv instead of jsonschema for validation

srijs:
Follow up from #161.

This is a breaking change again due to mapping through the underlying validation errors.


jsdevel:
I appreciate how thorough you are!  Not worried about the coverage.

PR #165, opened 2018-08-10
express-openapi: fix type definitions for ajv error

srijs:
This is to make sure we don't forget to change the type definitions for `express-openapi` when we cascade the ajv changes. Feel free to cherry-pick this commit into your own branch when you bump the versions!


jsdevel:
Thanks again @srijs !

jsdevel:
published as `v2.0.0`

PR #168, opened 2018-08-10
express-openapi: Bumping openapi-framework to 0.2.0 for ajv performance.

jsdevel:

PR #171, opened 2018-08-10
openapi-response-validator: basic support for OpenAPI 3 data types

srijs:
Context: https://github.com/kogosoftwarellc/open-api/pull/168#issuecomment-411954520

### DIFF
--- a/packages/openapi-response-validator/index.js
+++ b/packages/openapi-response-validator/index.js
@@ -1,6 +1,10 @@
 var Ajv = require('ajv');
 var LOCAL_DEFINITION_REGEX = /^#\/([^\/]+)\/([^\/]+)$/;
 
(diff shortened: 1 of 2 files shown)


srijs:
Rebased.

srijs:
Re-read the OpenAPI spec, and it says:

> [...], to support documentation needs, the format property is an open string-valued property, and can have any value. Formats such as "email", "uuid", and so on, MAY be used even though undefined by this specification.

Based on this, I don't think we should fail on unknown formats at all.

jsdevel:
ahh, nice catch!  In the event we _do_ decide to enforce formats at some point see https://github.com/kogosoftwarellc/open-api/pull/168#issuecomment-411958682

PR #201, opened 2018-09-24
fix doc to match index.js

naugtur:


jsdevel:
thanks!

naugtur:
I did it for myself ;) I'm already using this in 2 projects and if I start another one in a month I'll probably forget why openapi3 didn't work out of the box.

BTW, feature suggestion: read the version from the definition itself and default to that if available when no version passed in configuration. 

I'd selfishly PR that too, but I'm committed to using v3 only, so hardcoding the configuration works for me. ;)

Thanks for sharing your work!

PR #203, opened 2018-09-25
openapi-security-handler: Moving to ts and generalizing logic (fixes #197)

### LINKED ISSUE #197: Can't get security handler to return a 401 without a "challenge" attributes
**Context:**
Looking to return `401`s when a token is missing from a header. Front end redirects to login if this is the case

**Issue:**
The docs indicate that you can use [`cb(null, false)`](https://github.com/kogosoftwarellc/open-api/tree/master/packages/openapi-security-handler#response) from all securityHandlers to return a `401`. I only have one but I'm getting a `500` with:
```
No security handlers returned an acceptable response: token
```

**Notes:**
Looking at the code, it seems that without [`lastError.status`](https://github.com/kogosoftwarellc/open-api/blob/master/packages/openapi-security-handler/index.js#L99) defined there's no way to return a `401`

Also, when running `cb(null, false)` it seems the `lastError` becomes the first argument. So with `null` it would jump [`if (lastError)`](https://github.com/kogosoftwarellc/open-api/blob/master/packages/openapi-security-handler/index.js#L94) and it [the default `cb` block](https://github.com/kogosoftwarellc/open-api/blob/master/packages/openapi-security-handler/index.js#L135)

I'd be down to put in a PR, but I'm curious about you'd like the code to behave

Love your library! :two_hearts:

jsdevel:
* No longer requiring challenge or message in errors.
* Removing async package.
* Moving to promises instead of callbacks.


notxt:
totally! I can look at this later this afternoon!

jsdevel:
ping @notxt

notxt:
sorry! - ( it's like our big company meeting week where everyone flies in and goes out after work :beers: ) - I should be able to pull it down and test it after sprint planning

I'll try to get a couple of my coworkers to look over it too

question: are you worried about the coveralls fail? the color red is a strong motivator for me :stuck_out_tongue_winking_eye:

jsdevel:
> are you worried about the coveralls fail?

nope.  what happened was that i reduced the number of loc which brought the coverage down for the project as a whole.  coverage for the security handler remains at 100%.

notxt:
bah, my afternoon got all ate up and this evening is booked :grimacing: 

tomorrow! I'm wfh friday/monday so should be able to find some time! Sorry about the lag!

jsdevel:
@notxt i'm probably going to merge this to make progress in other areas.  probably not going to publish just yet though.  let me know what you think after reviewing.

PR #222, opened 2018-10-05
openapi-response-validator: now accepts nullable property + test

funduck:
test case shows error and fails but shouldn't in openapi v3

### DIFF
--- a/packages/openapi-response-validator/index.ts
+++ b/packages/openapi-response-validator/index.ts
@@ -119,7 +119,7 @@ function compileValidators(v, schemas) {
   const validators = {};
 
   Object.keys(schemas).forEach(name => {
-    validators[name] = v.compile(schemas[name]);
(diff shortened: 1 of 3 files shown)


jsdevel:
i may be able to fix this tonight

funduck:
I'm a bit confused, did I have to add test case that passes, but doesn't show the problem described in issue or I did right that wrote test case that fails and this is how issue is shown?

jsdevel:
@funduck sorry for the confusion.  This shows how to recreate the issue.  I've been pondering what the right way is to solve this problem.  I'm hoping to have a fix soon.

funduck:
@jsdevel is there something I should do for this pull request to be accepted?

jsdevel:
hey!  sorry i missed this.  i'll take a look tonight.

jsdevel:
Thanks @funduck !  I'll publish soon

PR #258, opened 2018-11-18
Support consumes MIME types from OpenAPI requestBody object

koocbor:
In order to use middlewear the consumes array must be populated with the mime-types an operation handles. The OpenAPI 3.0 specification moved the consumes definition to be the keys in the `requestBody.content` object. [https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#requestBodyObject](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#requestBodyObject)

This PR adds support to parse the `requestBody.content` keys to populate the consumes array so any defined middlewear for a mimetype is used.

I'm not sure if the plan is to allow piecemeal support of the OpenAPI 3.0 specification. This only addresses a small part of full support. I'm working on a project that is using the 3.0 spec and plan to add support as I run into issues. 

If the desire is for a more all-encompassing PR, that's probably beyond the scope of what I'll be able to do at this time.


packages/openapi-framework/index.ts
```diff
@@ -222,8 +222,13 @@ export default class OpenAPIFramework implements IOpenAPIFramework {
         const operationHandler = pathModule[methodName];
         methodName = METHOD_ALIASES[methodName];
         const operationDoc = handleYaml(getMethodDoc(operationHandler)) || pathDoc[methodName];
+        // consumes is defined as property of each operation or entire document
+        // in Swagger 2.0. For OpenAPI 3.0 consumes mime types are defined as the
+        // key value(s) for each operation requestBody.content object.
         const consumes = operationDoc && Array.isArray(operationDoc.consumes) ?
           operationDoc.consumes :
+            operationDoc && operationDoc.requestBody && operationDoc.requestBody.content ?
+            Object.keys(operationDoc.requestBody.content) :
```
jsdevel: does this need to exclude `default`?

koocbor:
I don't think so. I think I'm reading the spec correctly which says the
requestBody must have a map of mediatypes to mediatype objects. There is no
default defined.

Let me know if I'm missing the point of your question.

Thanks,
Rob

jsdevel:
thanks!

jsdevel:
published as follows:

```
openapi-framework@0.9.0
express-openapi@3.2.0
koa-openapi@3.1.0
```

PR #259, opened 2018-11-18
Added

jwmur:
1) support for v3 security handlers and
2) a basepath argument for v3 schemas

### DIFF
--- a/packages/openapi-framework/index.ts
+++ b/packages/openapi-framework/index.ts
@@ -107,7 +107,7 @@ export default class OpenAPIFramework implements IOpenAPIFramework {
     this.enableObjectCoercion = !!args.enableObjectCoercion;
     this.originalApiDoc = handleYaml(handleFilePath(args.apiDoc));
     this.apiDoc = copy(this.originalApiDoc);
-    this.basePath =  (this.apiDoc.basePath || '').replace(/\/$/, '');
+    this.basePath =  (this.apiDoc.basePath || args.basePath || '').replace(/\/$/, '');
     this.validateApiDoc = 'validateApiDoc' in args ?
         !!args.validateApiDoc :
         true;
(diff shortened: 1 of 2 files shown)


jsdevel:
thanks!  can you add some tests?

jsdevel:
@jwmur you still working this?  going to close soon without tests.

jwmur:
Yes.  Any tips on where similar tests?

jsdevel:
any progress @jwmur ?

jwmur:
Carving out time to write the tests.  I'm still planning on doing.

jwmur:
Will try to get done before the weekend

jsdevel:
Thanks!  also, can you update the commit subject?  i want to be able to track this in git history.

jsdevel:
@bill-kitsune are there ideas here that are relevant to #274 ?

@jwmur ETA on tests?  Also looks like there's conflicts.

bill-kitsune:
@jsdevel I think this approach to the baseUrl in this PR would get people by until #274 is resolved, but I guess doesn't quite "Support the servers attribute" of OpenAPI 3.0.  It seems more like an approach to get people by until the framework supports the servers attribute.

The goal of this library also seems like as much as possible should be derived from the openapi doc, to that end it might be better to split this PR to just the security handlers and then wait for #274 ?

bill-kitsune:
In express at least a way to work around the base path issue until #274 might be something like the following:

```
const apiRouter = express.Router()
expressOpenApi.initialize({
  apiDoc: v1ApiDoc,
  app: apiRouter,
  dependencies: {
  worldsService: WorldsService,
  },
})
this.express.use('/api/v1', apiRouter)
```

jsdevel:
@bill-kitsune I agree.  @jwmur can you limit this PR to the securityHandlers?

jsdevel:
@jwmur: @bill-kitsune has just added support for the servers attribute, we'd love to get your work on the securityHandlers in.  Can you limit this PR to just that?

jsdevel:
@jwmur the work on securityHandlers herein could potentially fix #305.  Are you going to finish this soon?

jsdevel:
@jwmur going to close this due to a lack of response.  we'd love to see support for security handlers, so if you have time to address the issues stated above, please re-open!

Issue #269, opened 2018-12-11
Question: Does openapi-response-validator support v3 schemas?


dosyara:
I'm doing some tests with OpenAPI v3 and cannot figure out how to validate response object with `validateResponse`.
I use:
"express": "^4.16.3",
"express-openapi": "^3.3.0",

I have `openapi: "3.0.0"` schema yalm file. File itself is valid according to v3 validator.
Paths section looks like this:
```
paths:
  /test:
    get:
      responses:
        "200":
          description: test path
          content:
            application/json:
              schema:
                type: object
                  required:
                    - id
                    - date
                  properties:
                    id:
                      type: string
                    date:
                      type: string
```

When I run 
```js
res.validateResponse(statusCode, responseObject)
```
for a valid response object, I get:
```js
{ message: 'The response was not valid.',
  errors:
   [ { errorCode: 'type.openapi.responseValidation',
       message: 'response should be null' } ] }
```

So my question is: is this happens because of lack of support by response validator or am I doing something wrong?

jsdevel:
@dosyara this may have been fixed by @funduck in https://github.com/kogosoftwarellc/open-api/pull/222.  I just published it.  Can you try it out?

dosyara:
Tried, no luck.
 Updated express-openapi to 3.4.0, and still get `response should be null`.

jsdevel:
what arguments are you passing to validateResponse?

dosyara:
I pass: two parameters: 200 (`statusCode`) and { my json response } ( `responseObject`).
I've experimented with different response objects, the only thing which not generates error is `null`

funduck:
pushed to #224, hope I did it right and @jsdevel will soon accept my request

jsdevel:
you did @funduck !  try with the latest @dosyara

dosyara:
Updated to express-openapi@3.5.0 it now works with inlined schemas!
But have an issue with references. I get an error if schema is defined in components:
```
{ message: 'The response was not valid.',
  errors:
   [ { errorCode: '$ref.openapi.responseValidation',
       message:
        'response can\'t resolve reference #/components/schemas/test' } ] }
```
For this schema:
```
openapi: "3.0.0"
paths:
  /test:
    get:
      parameters:
        - name: id
          in: query
          required: true
          schema:
            type: string
      responses:
        "200":
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/test'

components:
  schemas:
    test:
      type: object
[...]

PR #271, opened 2018-12-15
fix: Fix OpenAPI v3 response validation for schemas with references

dosyara:
Following discussion in #269, creating this PR.
Fixed an issue with openapi schemas with references.

I believe fix is slightly sloppy, because it mixes v2 and v3 specific fields in code. 
Would be nice to refactor code, so it would be easier to implement specific behaviour for specific version of the schema.


jsdevel:
Looks good!  Thank you so much!

jsdevel:
published in latest versions

Issue #162, opened 2018-08-09
Support OpenApi 3.0 requestBody


seeker89:
Hi and thanks for this package !

I've noticed, that https://github.com/kogosoftwarellc/open-api/tree/master/packages/openapi-request-validator doesn't yet support `requestBody`, so if provided with OpenApi 3.0 spec, things mostly work, apart from validating the actual body.

Are there plans to add support for that ?

Thanks !

jsdevel:
Right now I'm looking for the community to provide support for OpenAPI v3.  If you have the time to open a PR to add that functionality, it would be very much appreciated.

seeker89:
Gotcha. Thanks for your answer.

jsdevel:
closed by @bill-kitsune in #272!

PR #272, opened 2018-12-18
Request validator request body

bill-kitsune:
This is a step toward resolving #162 

My approach was to try to keep it backwards compatible, but I imagine that there could be a couple of different approaches to handling v3, so I'm hoping this at least starts the conversation.

### DIFF
--- a/packages/openapi-request-validator/README.md
+++ b/packages/openapi-request-validator/README.md
@@ -27,6 +27,17 @@ var validator = new OpenAPIRequestValidator({
       required: true
     }
   ],
+  requestBody: { // optional OpenApi v3 requestBodyObject
+    contents: {
+      'application/json': {
+        schema: {
+          properties:
+            name:
+              type: 'string'
+        }
+      }
+    }
+  },
   schemas: null, // an optional array or object of jsonschemas used to dereference $ref
   version: 'swagger-2.0', // default optional value for future versions of openapi
   errorTransformer: null, // an optional transformer function to format errors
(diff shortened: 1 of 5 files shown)


jsdevel:
nice work!  i'll publish later tonight

jsdevel:
published in latest versions

mbsimonovic:
this works fine for numbers but seems like it doesn't validate string's minimum and maximum lengths? Here's a demo:
```
const OpenAPIRequestValidator = require('openapi-request-validator');

const requestBodySpec ={
    "description": "request body",
    "required": true,
    "content": {
        "application/json": {
            "schema": {
                "type": "object",
                "required": [
                    "query"
                ],
                "properties": {
                    "query": {
                        "type": "string",
                        "minimum": 2,
                        "maximum": 1024
                    },
                    "maxResults": {
                        "description": "maximum entries",
                        "type": "integer",
                        "minimum": 1,
                        "default": 1000
                    }
                }
            }
        }
    }
};

var request = {
    headers: {
        'content-type':
            'application/json'
    },
    body: {
        query: 'a', // should fail, minimum length violated but DOESNT FAIL!?
        // query: new Array(150 + 1).join('1234567890'), // DOESNT FAIL!?
        // maxResults: 'a', // fails
    }
};

var errors = validator.validate(request);
console.log(errors); // => null, it's ok for foo to exist, and name is not required

```

Is this intentional or should I create a new issue?

bill-kitsune:
@mbsimonovic not intentional just possibly incomplete.  I would open up a separate issue for this.

mbsimonovic:
@bill-kitsune sure, https://github.com/kogosoftwarellc/open-api/issues/287

mbsimonovic:
turned out to be a typo 😤

Issue #273, opened 2018-12-18
Thoughts on using a linter/formatter for this project? 


bill-kitsune:
I was working on a PR and noticed a lot of the code was written with a certain style, that I found tedious to try to match.  

Would the maintainers be interested in or have any plans to add a linter to the project? Prettier, or even just tslint?

jsdevel:
> Would the maintainers be interested in or have any plans to add a linter to the project?

yes!

> Prettier, or even just tslint?

both are fine, but no git hooks please

jsdevel:
beautifully closed by #275!  thanks again @bill-kitsune !

PR #275, opened 2018-12-19
Linter

bill-kitsune:
#273 

Uses tslint:recommended, and prettier for formatting.  

It didn't seem like there was too much emphasis on some of the things that tslint & prettier fix before so I will understand if no one's that into it, but I always feel like have a slightly more consistent style improves readability.  tslint does help catch some things that can be gotchas as well. 

Most IDE's now have plugins to automatically run this stuff and fix it as you work anyways which at that point makes it super trivial to follow the rules.

p.s. It's a wonderful thing that your libraries are so well tested.

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,0 +1,3 @@
+{
+  "singleQuote": true
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ notifications:
   email:
     on_success: "never"
 script:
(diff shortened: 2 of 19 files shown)


jsdevel:
@bill-kitsune this is FANTASTIC!  Thank you so much!

Quick question, will the build fail if prettier linting is off?  My hope is yes.

jsdevel:
verified!  this was huge!  thank you so much!

PR #276, opened 2018-12-20
Linter (again)

bill-kitsune:
Sorry for the double PR but after you merged in the last one I realized there was an issue with the glob pattern I was using.  The issue was that I was allowing it to use the shell glob which is not recursive. So I ended up missing any files that were in a sub-directory of a package.  I.e. if a package has a src, or test directory.

Changing  ` "lint": "tslint ./packages/**/*.[j,t]s",` to `"lint": "tslint './packages/**/*.[j,t]s'",` solved that an opened up a much larger can of unlinted files for me to tackle... 

A couple of other things that are good to be aware of:
The follow section in `tslint.json` controls which files to ignore.  So if there are any I've missed please let me know, or add them as needed. 

```"linterOptions": {
    "exclude": [
      "**/node_modules/**/*",
      "**/coverage/**/*",
      "**/.nyc_output/**/*",
      "**/dist/**/*",
      "**/docs/**/*",
      "packages/fetch-openapi/test/**/output.js"
    ]
  },
```

Without re-writing fetch-openapi, or running it's output through prettier it seemed easiest to just ignore the output files for now, I didn't want to add more than was necessary to the scope of this PR because there is already a lot of files changed.


jsdevel:
> Without re-writing fetch-openapi, or running it's output through prettier it seemed easiest to just ignore the output files for now, I didn't want to add more than was necessary to the scope of this PR because there is already a lot of files changed.

yea, that package should get re-written at some point to rely on `fn.toString` instead of concatenating strings (problem for a different day).

jsdevel:
you've been kicking some major butt chuck norris style @bill-kitsune !  thank you again!

PR #277, opened 2018-12-21
express-openapi: Use OpenApi 3.0 servers attribute to build baseUrls

### LINKED ISSUE #274: Support for servers attribute for OpenAPI 3.0
2.0 uses a basePath attribute that the library uses to modify the paths that are created.  3.0 uses the servers attribute which is a list of servers.

Initial thoughts on how this could work is something like the following:

Add a useServer argument to `OpenAPIFrameworkArgs` interface. 

The arg could have the following behavior:

* Is optional
* Only matters if the apiDoc is a OpenAPI 3.x.x document
* If not set the framework will use the first servers path for basePath
* If set the framework will match the argument to the server description
* If set and cannot match on server description throw error (thoughts on this?)

This would allow people to do the following:

```
expressOpenApi.initialize({
      apiDoc,
      app: this.express,
      dependencies: {
        worldsService: WorldsService,
      },
      useServer: process.env.NODE_ENV || 'development'
    })
```

Thoughts?

bill-kitsune:
Closes #274

Turns out I didn't really need  to touch `openapi-framework` at all.  A lot of what I wanted to do was rather express specific, and the ctx objects already expose the apiDoc.

### DIFF
--- a/packages/express-openapi/index.ts
+++ b/packages/express-openapi/index.ts
@@ -1,13 +1,14 @@
 import { ErrorObject, FormatDefinition, FormatValidator } from 'ajv';
 import { Application, ErrorRequestHandler, RequestHandler } from 'express';
 import OpenAPIFramework, {
+  BasePath,
   OpenAPIFrameworkArgs,
   OpenAPIFrameworkConstructorArgs
 } from 'openapi-framework';
 import { OpenAPIRequestValidatorError } from 'openapi-request-validator';
 import { OpenAPIResponseValidatorError } from 'openapi-response-validator';
 import { SecurityHandlers } from 'openapi-security-handler';
-import { OpenAPI } from 'openapi-types';
+import { OpenAPI, OpenAPIV3 } from 'openapi-types';
 const CASE_SENSITIVE_PARAM_PROPERTY = 'x-express-openapi-case-sensitive';
 const normalizeQueryParamsMiddleware = require('express-normalize-query-params-middleware');
 const loggingPrefix = 'express-openapi';
@@ -81,23 +82,29 @@ export function initialize(args: ExpressOpenAPIArgs): OpenAPIFramework {
   framework.initialize({
     visitApi: ctx => {
       if (exposeApiDocs) {
-        // Swagger UI support
-        app.get(ctx.basePath + docsPath, (req, res, next) => {
-          // @ts-ignore
-          req.apiDoc = ctx.getApiDoc();
-          // @ts-ignore
-          if (req.apiDoc.swagger) {
-            // @ts-ignore
-            req.apiDoc.host = req.headers.host;
-            // @ts-ignore
-            req.apiDoc.basePath = req.baseUrl + ctx.basePath;
-          }
-          securityFilter(req, res, next);
-        });
-      }
+        const basePaths = [];
+        const apiDoc = ctx.getApiDoc();
+        basePaths.push(...ctx.basePaths.map(toExpressBasePath));
 
-      if (errorMiddleware) {
-        app.use(ctx.basePath, errorMiddleware);
+        // Swagger UI support
+        for (const basePath of basePaths) {
+          app.get(basePath + docsPath, (req, res, next) => {
+            // @ts-ignore
+            req.apiDoc = ctx.getApiDoc();
+            // @ts-ignore
+            if (req.apiDoc.swagger) {
+              // @ts-ignore
+              req.apiDoc.host = req.headers.host;
+              // @ts-ignore
+              req.apiDoc.basePath = req.baseUrl + basePath;
+            }
+            securityFilter(req, res, next);
+          });
+
+          if (errorMiddleware) {
+            app.use(basePath, errorMiddleware);
+          }
+        }
       }
     },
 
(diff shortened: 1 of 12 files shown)


packages/express-openapi/test/sample-projects/with-servers/spec.js
```diff
@@ -0,0 +1,141 @@
...
+        const app = express();
+        openapi.initialize({
+          app,
+          apiDoc: generateOpenApiDocWithServers(test.servers),
+          paths: path.resolve(__dirname, 'api-routes')
+        });
+        request = supertest(app);
+      });
+      for (let i of test.expectedPaths) {
+        it(`route should be available at ${i}`, async function() {
```
jsdevel: can you use `` it(`should ``...`?

packages/express-openapi/index.ts
```diff
@@ -255,3 +275,25 @@ function toPromiseCompatibleMiddleware(fn) {
   }
   return fn;
 }
+
+function getBasePathsFromServers(servers: OpenAPIV3.ServerObject[]): string[] {
```
jsdevel: i kind of feel like this _should_ be in `openapi-framework`.  this logic doesn't seem specific to express, and `koa-openapi` would benefit from being passed the basePaths.  Perhaps `openapi-framework` should pass `basePaths` as an array instead of `basePath` as a string?

jsdevel:
i think the build for v6 is failing cause async/await wasn't added until node v7.  can you remove async/await?  until v6 is out of maintenance support (https://nodejs.org/en/about/releases/) we should probably continue to support it.

packages/express-openapi/test/sample-projects/with-servers/api-routes/users.js
```diff
@@ -0,0 +1,10 @@
+var Promise = require('bluebird');
```
jsdevel: is it needed to require bluebird here?

packages/express-openapi/index.ts
bill-kitsune: That is where I started with it.  I was going to expose an `baseUrls` attribute or something from the `ctx` object.  I thought to move it into express specifically because I was starting to convert the OpenApi param syntax `/{var}` to express specific syntax `/:var`.  

I guess `koa-router` also uses express routing syntax, but then `openapi-framework` would be kind of glued to that style of routing.  For example I'm not sure what if there are any plans to support hapi.js at all but they use path params like: `'/hello/{user}'`

Although I guess that being said OpenApi picked one way, we converted it to express style, if `openapi-framework` exposes express style no reason someone couldn't just convert them to whatever they needed.

jsdevel: yea, i think it's find if openapi-framework exposes the urls in one format or another, but i do feel that baseUrls should be exposed there.

packages/express-openapi/test/sample-projects/with-servers/api-routes/users.js
bill-kitsune: Nope, removing.

packages/express-openapi/index.ts
bill-kitsune: Okay, I'll switch it around and re-push. I'll also add the same logic to `koa-openapi`

packages/express-openapi/test/sample-projects/with-servers/spec.js
bill-kitsune: Reworked the mocha description a bit so it hopefully reads better

packages/express-openapi/index.ts
jsdevel: sounds good!  if you link locally to get `(?:express|koa)-openapi` working with `openapi-framework`, i can handle the publishing.

bill-kitsune:
@jsdevel also let me know if you want me to break out the project specific merge requests.  It might be easier to see/validate what has changed in what?

packages/openapi-framework/src/types.ts
```diff
@@ -70,12 +70,14 @@ interface OpenAPIFrameworkArgs {
 
 export interface OpenAPIFrameworkAPIContext {
   basePath: string;
+  basePaths: BasePath[];
```
jsdevel: can we just use `basePaths`?  We can always iterate over that for V2 in downstream projects like `(?:express|koa)-openapi` even if there's only 1 basePath in the array.

bill-kitsune: Yeah I just wasn't sure if that would be cool since it would be a breaking change, but I suppose that is what semver is meant to help with.  I will try to re-push at some point today.

jsdevel: right now `openapi-framework` hasn't had a major version, so we can make breaking changes to it's API without forcing breaking changes on downstream projects.

bill-kitsune:
@jsdevel latest changes include switching over to basePaths.

jsdevel:
@bill-kitsune everything looks great.  the only issue is the build.  i was hoping i'd have a link script working to avoid the need to submit multiple PRs, but had some issues with it (you can see it [here](https://github.com/kogosoftwarellc/open-api/pull/284) (i'm starting to think adopting lerna may be best but i can't get over the duplication with that)).  can you create a new pr with the koa/express changes and limit this to `openapi-framework`?  that way i can get that published.

bill-kitsune:
@jsdevel for sure.  Honestly it will probably be easier that way.  I don't mind the separate PR's that much since they are separate projects, and then the commits will be cleaner and can use the prefix thing.

bill-kitsune:
@jsdevel gonna close this, and will make the separate PR's.

Issue #279, opened 2018-12-31
fails to handle 'multipart/form-data' in openapi v3 (Unsupported Content-Type multipart/form-data; boundary=...)


mbsimonovic:
i'm getting an error trying to handle file uploads with v3.6.0. The spec is:

```
openapi: "3.0.0"
servers:
    - url: http://localhost:3000
      description: local server

paths:
    /upload/:
        post:
            summary: Create NEW files
            operationId: upload
            requestBody:
                required: true
                content:
                    multipart/form-data:
                        schema:
                            type: object
                            properties:
                                docs:
                                    type: array
                                    items:
                                        type: string
                                        format: binary
# tried this as well but also doesnt work:
                        # schema:
                        #   type: object
                        #   properties:
                        #     upload:
                        #       type: string
                        #       format: binary

```
 express-openapi is initialised as suggested:
```
 'multipart/form-data'(req, res, next) {
            multer().any()(req, res, (err) => {
                if (err) return next(err);
                req.files.forEach(f => req.body[f.fieldname] = f);
                return next(); // this executes correctly
            });
        }
```

Hitting the app with curl or swagger-ui:
`curl   "http://localhost:3000/upload/" -H "accept: */*" -H "Content-Type: multipart/form-data" -F docs=@package.json ` fails with the following error (the endpoint never gets called):

```
Unsupported Content-Type multipart/form-data; boundary=----WebKitFormBoundaryWyK9kAU7d35AKf26
```

The same happens with superagent/supertest and postman. Any hints?

mbsimonovic:
`function getSchemaForMediaType(contentType:'multipart/form-data; boundary=--------------------------480295105673604652952056', requestBodySpec='content: multipart/form-data') {` seems to be the problem:

<img width="1660" alt="screen shot 2018-12-31 at 09 46 54" src="https://user-images.githubusercontent.com/888008/50557023-1fab3680-0ce1-11e9-94c6-79861d8692f9.png">

jsdevel:
@mbsimonovic please submit a PR to resolve this issue.

mbsimonovic:
@jsdevel not sure now to setup and run tests, but the simplest fix could look like this (line 313):

```javascript
if (contentType.startsWith('multipart/form-data') && mediaTypeKey.startsWith('multipart/form-data')) {
               return mediaTypeKey;
}
```

mbsimonovic:
another option is to use a module, https://github.com/jshttp/content-type for example.

[...]

PR #285, opened 2019-01-03
openapi-request-validator: bugfix: handle multipart/form-data (https:…

mbsimonovic:
…//github.com/kogosoftwarellc/open-api/issues/280)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,8 @@ dist
 *.swp
 *.swo
 packages/*/dist
+# tools artifacts
+.idea
+.vscode
+*~
+.DS_Store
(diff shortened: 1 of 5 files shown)


mbsimonovic:
hm, is it possible that contentType is undefined? in that case the parser will fail and throw an error.

mbsimonovic:
ah, linter fails..

jsdevel:
this is great!  thank you so much @mbsimonovic !

mbsimonovic:
shall we close https://github.com/kogosoftwarellc/open-api/issues/279 ?

jsdevel:
does this close #279?

mbsimonovic:
yes, with the pull request it will properly handle multipart

PR #293, opened 2019-01-05
openapi-framework: Resolve response and parameter references for OpenAPI 3.0

### LINKED ISSUE #289: Resolve response references for OpenAPI 3.0 (#/components/responses instead of #/responses)
# The problem

When defining reusable responses within the components object of the OpenAPI Document as specified in the [Open API 3.0.2 specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#components-object) the schema validation fails for Open API version > 3.0.0. The response references for V2.0 are resolved well.

Affected package: openapi-framework (util function resolveResponseRefs)

# Example

Following example spec fails when using the express-openapi middleware.

> express-openapi: Invalid response $ref or definition not found in apiDoc.responses: #/components/responses/Error

## The Open API Spec
```json
{
  "openapi": "3.0.2",
  "info": {
    "version": "2.0.0",
    "title": "Basic Usage Example"
  },
  "paths": {},
  "components": {
    "responses": {
        "Error": {
          "description": "An error occurred",
          "content": {
            "application/json": {
              "schema": {
                "type": "object",
                "required": [
                  "error"
                ],
                "properties": {
                  "error": {
                    "type": "object",
                    "required": [
                      "message"
                    ],
                    "properties": {
                      "message": {
                        "type": "string"
                      }
                    }
                  }
                }
              }
            }
          }
        }
      }
  }
}
```

## Path Handler

```javascript
/*******************************************************************
 *   paths/config.js
 *******************************************************************/
module.exports = function (config) {
     let operations = {
          GET
     };

     operations.GET.apiDoc = {
          summary: 'Returns config.',
          operationId: 'getConfig',
          parameters: [],
          responses: {
               200: {
                    description: 'Success',
                    schema: {
                         $ref: '#/components/schemas/Config'
                    }
               },
               default: {
                    $ref: '#/components/responses/Error'
               }
          }
     };

     function GET(req, res, next) {
          res.validateAndSend(200, config);
     }

     return operations;
};
```
# Affected Open API util function

The resolveResponseRefs function tries to get the definition from apiDoc.responses (#/responses) instead of apiDoc.components.responses (#/components/responses).

```javascript
/*******************************************************************
 *   /node_modules/openapi-framework/dist/src/util.js
 *******************************************************************/
function resolveResponseRefs(framework, responses, apiDoc, route) {
    return Object.keys(responses).reduce(function (resolvedResponses, responseCode) {
        var response = responses[responseCode];
        if (typeof response.$ref === 'string') {
            var match = RESPONSE_REF_REGEX.exec(response.$ref);
            var definition = match && (apiDoc.responses || {})[match[1]];
            if (!definition) {
                throw new Error(framework.name + ": Invalid response $ref or definition not found in apiDoc.responses: " + response.$ref);
            }
            resolvedResponses[responseCode] = definition;
        }
        else {
            resolvedResponses[responseCode] = response;
        }
        return resolvedResponses;
    }, {});
}
```

My suggestion would be to differentiate between the different API versions and handle the resolving accordingly:

```javascript
var apiVersion = apiDoc.swagger ? apiDoc.swagger : apiDoc.openapi;
var responses = apiVersion === "2.0" ? apiDoc.responses : apiDoc.components.responses;
```

christian-knecht:
Fix #289 and not documented issue.


packages/openapi-framework/test/sample-projects/paths-dir-with-responses-defined-as-specified-in-openapi-3/spec.ts
```diff
@@ -0,0 +1,25 @@
...
+  beforeEach(() => {
+    framework = new OpenapiFramework({
+      apiDoc: path.resolve(__dirname, 'apiDoc.yml'),
+      featureType: 'middleware',
+      name: 'some-framework',
+      paths: path.resolve(__dirname, 'paths')
+    });
+  });
+
+  it('#/components/responses/Error should be resolved properly', () => {
```
jsdevel: can you change this to `it('should be resolved properly...`?

jsdevel:
thanks @christian-knecht !

PR #295, opened 2019-01-07
openapi-framework: OpenApi v3 servers support

bill-kitsune:
OpenAPI v3 support for openapi-framework.  Note that I'm replacing basePath with basePaths so this will be a breaking change.

After this is published I can make the `express-openapi` and `koa-openapi` pull requests.


bill-kitsune:
@jsdevel not sure why the node v6 build is failing seems like a weird error to me, and I cannot reproduce it locally.

bill-kitsune:
> @jsdevel not sure why the node v6 build is failing seems like a weird error to me, and I cannot reproduce it locally.

I think I found it.  I think my IDE automatically added an import that didn't really make sense.

jsdevel:
this is great!  thank you so much @bill-kitsune !

don't worry about this being a breaking change, according to semver anything <1 has no guarantees to stability.  openapi-framework should be referenced by the exact version for the time being so we're good.

Issue #294, opened 2019-01-07
v3: doesnt validate requestBody when no parameters are defined


mbsimonovic:
For operations with no parameters but only request body, there's no request validation. Here's an example:

```
    /somepath:
        post:
            requestBody:
                content:
                    application/json:
                        schema:
                            type: object
                            properties:
                                name:
                                    type: string
                                    minLength: 1

            responses:
                '201':
                    description: ok created
```
The code responsible for setting up request validation:

```
                        var methodParameters = util_1.withNoDuplicates(util_1.resolveParameterRefs(_this, Array.isArray(operationDoc.parameters)
                            ? pathParameters.concat(operationDoc.parameters)
                            : pathParameters, parameterDefinitions));
                        operationContext.methodParameters = methodParameters;  // EMPTY ARRAY
                        if (methodParameters.length) {
                            // defaults, coercion, and parameter validation middleware
                            if (util_1.allowsValidationFeature(_this, _this.apiDoc, pathModule, pathDoc, operationDoc)) {
                                var requestValidator = new openapi_request_validator_1["default"]({
                                    errorTransformer: _this.errorTransformer,
                                    parameters: methodParameters,
                                    schemas: _this.apiDoc.definitions,
                                    externalSchemas: _this.externalSchemas,
                                    customFormats: _this.customFormats,
                                    requestBody: operationDoc.requestBody
                                });
                                operationContext.features.requestValidator = requestValidator;
                            }
```

`methodParameters` is empty, so request validator is never created. To fix it it's enough to add `|| operationDoc.requestBody` to the if condition:
```
                        if (methodParameters.length || operationDoc.requestBody) {
// instead of:
//                        if (methodParameters.length) {

```

I'd submit a PR if this is ok?

Tested on v3.7.0, openapi v3, node v10.

jsdevel:
yes, please submit a PR.  also, i have a few modules that i need to publish tonight, so it's possible that this has already been fixed.

PR #297, opened 2019-01-08
Bugfix for issue #294

mbsimonovic:
https://github.com/kogosoftwarellc/open-api/issues/294

### DIFF
--- a/packages/openapi-framework/index.ts
+++ b/packages/openapi-framework/index.ts
@@ -362,7 +362,7 @@ export default class OpenAPIFramework implements IOpenAPIFramework {
               );
               operationContext.methodParameters = methodParameters;
 
-              if (methodParameters.length) {
+              if (methodParameters.length || operationDoc.requestBody) {
                 // defaults, coercion, and parameter validation middleware
                 if (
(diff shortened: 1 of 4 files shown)


jsdevel:
the commits look kind of odd.  why are there release commits herein?

mbsimonovic:
i've only changed the commit messages. guess because i had to merge?

jsdevel:
commits

ideally, i'd like to see 1 commit here

mbsimonovic:
ah here's what happened, i wrote the test and committed, and then you pushed some changes, so i pulled to see if the bug got fixed, but wasn't, so i fixed the bug later. i'm not sure how to squash the first and fourth commit into a single one. would it be ok to accept it as it is now, and i'll try to make it a single commit in the future?

mbsimonovic:
i agree, in retrospect, i should've created another branch to pull your changes and test, before making the bugfix commit, mea culpa, lesson learned.

mbsimonovic:
it's a single commit now

jsdevel:
No worries!  this is great thank you!

PR #299, opened 2019-01-08
Express koa openapiv3 servers

### LINKED ISSUE #274: Support for servers attribute for OpenAPI 3.0

bill-kitsune:
`express-openapi` and `koa-openapi` now take advantage of new basePaths attribute from `openapi-framework` which is derived from either the basePath attribute in openapi v2 or the servers attribute from openapi v3.  

Closes #274

### DIFF
--- a/packages/express-openapi/index.ts
+++ b/packages/express-openapi/index.ts
@@ -1,13 +1,14 @@
 import { ErrorObject, FormatDefinition, FormatValidator } from 'ajv';
 import { Application, ErrorRequestHandler, RequestHandler } from 'express';
 import OpenAPIFramework, {
+  BasePath,
   OpenAPIFrameworkArgs,
   OpenAPIFrameworkConstructorArgs
 } from 'openapi-framework';
 import { OpenAPIRequestValidatorError } from 'openapi-request-validator';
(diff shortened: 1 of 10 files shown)


jsdevel:
great job!  thanks @bill-kitsune !  i'll publish tonight.

mbsimonovic:
there's nothing in readme about this, how can i configure `/v1` as base path for openapi v3? setting servers.url to `http://localhost:3000/v1` or `/v1` doesn't work (i'm on express-openapi v4.0.0)

Issue #301, opened 2019-01-09
openapiv3 : missing validation for header, formdata and path parameters


maxime-morand:
All validation informations for headers, formdata and path parameters are ignored for an openapi 3 specification (except checking for mandatory parameters).
For example with this API :
```json
{
	"/myApi/{foo}": {
		"get": {
			"parameters": {
				"foo": {
					"name": "foo",
					"in": "path",
					"required": true,
					"schema": {
						"type": "string",
						"pattern": "^[a-z]{3}$"
					}
				},
				"bar": {
					"name": "bar",
					"in": "query",
					"schema": {
						"type": "string",
						"pattern": "^[a-z]{3}$"
					}
				}
			}
		}
	}
}
```
`GET /myApi/abc?bar=abcef` returns a validation error for the bar parameter
`GET /myApi/abcef?bar=abc` should return a validation error for the foo parameter

To my mind, all parameters should have the same validation process, whatever its type.

The problem seems to come from openapi-jsonschema-parameters, which, unlike for query parameters, only look for validation keywords at the root of the parameter definition and not in his shema property.

I can submit a PR if you are ok with that.

jsdevel:
yes, please submit a PR

PR #304, opened 2019-01-10
openapi-jsonschema-parameters : headers, formdata and path parameters validation for openapi3

maxime-morand:
Fix for #301


jsdevel:
thank you so much @maxime-morand !

Issue #280, opened 2018-12-31
can't resolve reference to a schema component for requestBody


mbsimonovic:
GET and POST reference the same schema component, GET works fine, but POST fails with the following error:

```
 {
      "message": "can't resolve reference #/components/schemas/Document",
      "location": "body",
      "schema": {
        "$ref": "#/components/schemas/Document"
      }
    }
```
The openapi-doc is as follows, and according to the specification https://swagger.io/docs/specification/describing-request-body/ it should be correct:

```yaml
openapi: "3.0.0"
paths:
    /path/{documentPath}:
        get:
            summary: Returns a document by path
            parameters:
                - name: documentPath
                  in: path
                  required: true
                  schema:
                      type: string
            responses:
                '200':
                    description: Success
                    content:
                        application/json:
                            schema:
                                $ref: "#/components/schemas/Document"
                default:
                    description: Error
                    content:
                        application/json:
                            schema:
                                $ref: "#/components/schemas/ErrorResponse"
        post:
            summary: Create a new document by path
            parameters:
                - name: documentPath
                  in: path
                  required: true
                  schema:
                      type: string
            # https://swagger.io/docs/specification/describing-request-body/
            requestBody:
                required: true
                content:
                    application/json:
                        schema:
                            $ref: "#/components/schemas/Document"
            responses:
                '201':
                    content:
                        application/json:
                            schema:
                                $ref: "#/components/schemas/Document"
                default:
                    description: Error
                    content:
                        application/json:
                            schema:
                                $ref: "#/components/schemas/ErrorResponse"
components:
    schemas:
        Document:
            type: object
            required:
                - path
                - type
                - uid
            properties:
                name:
                    type: string
                title:
                    type: string
                path:
                    type: string
                uid:
                    type: string
                type:
                    type: string
                    enum: ['Folder', 'File', 'Root', 'Picture', 'Domain']
                    default: 'File'
                properties:
                    type: object # key-value pairs
```

relativityboy:
I'm Running into the same issue, and I've triple checked it.
```
{ status: 400,
[0]   errors:
[0]    [ { message:
[0]         'can\'t resolve reference #/components/schemas/appToggleSetting',
[0]        location: 'body',
[0]        schema: [Object] } ] }

```

mbsimonovic:
the error happens because:
```
//OpenAPIRequestValidator.prototype.validate():
[...]

PR #309, opened 2019-01-11
second PR, openapi-framework: pass v3 schema definitions to request validator (c…

mbsimonovic:
…loses #280)


jsdevel:
thank you!

Issue #11, opened 2016-03-03
Option to validate all responses


DaleJefferson:
I can't find an option to validate all responses, it gets repetitive to keep doing.

```
const validationError = res.validateResponse(200, json);
if (validationError) {
    return next(validationError);
}

res.json(json);
```

jsdevel:
Yea, I agree.  What would you suggest?

jsdevel:
I usually add middleware to my stack that formats my data in a certain way.

So for example, I may call a service from my controller, and that service will give me most of what I need.  But usually in an API you have additional information you want to return.

Here's an example of what I may want to return from my API layer:

```javascript
{
  "statusCode": 200,
  "resource": {
    "name": "Tom Foolery"
  }
}
```

From my service layer though, I'm not concerned about status code, so my services would return this:

```javascript
{
  "name": "Tom Foolery"
}
```

My controller would then look like this:

```javascript
get: [addResourceResponseMiddleware, function(req, res) {
  myPersonService.getPerson(function(err, person) {
    res.sendAPIResource(200, person);
  });
}]
```

addResourceResponseMiddleware would look like this:

```javascript
function addResourceResponseMiddleware(statusCode, resource) {
  var proposedResponse = {
    statusCode: statusCode,
    resource: resource
  };
  const validationError = res.validateResponse(statusCode, proposedResponse);
  if (validationError) {
    return next(validationError);
  }

  res.status(statusCode).json(proposedResponse);
}
```

DaleJefferson:
Personally I found this option confusing,

> x-express-openapi-disable-response-validation-middleware

I assumed that response validation was turned on i.e. all responses were validated and that option opted you out for a specific request.

This would be the simplest and then you could use x-express-openapi-disable-response-validation-middleware to opt out.
```
app.set('response-validation', true);
```

But maybe this is a better idea?

```
res.validateAndSendJSON(200, json);
```

jsdevel:
> I assumed that response validation was turned on i.e. all responses were validated and that option opted you out for a specific request.

Yea, I can see that.  Perhaps that should be renamed to `x-express-openapi-add-validation-methods`?

jsdevel:
> res.validateAndSendJSON(200, json);

I like this.  What would it look like for other content types like XML?

jsdevel:
Almost seems like `res.validateAndSend` would be best, as the content type should be defined by your openapi document.

DaleJefferson:
[...]

PR #311, opened 2019-01-13
express-openapi: Add custom middleware examples to README.md

maxwellb:
Addresses #11 

Provide an example of adding a custom middleware in Express which runs on all paths and validates the response sent from the controller.


packages/express-openapi/README.md
```diff
@@ -368,6 +368,57 @@ available in the affected operation handler method.
 * `'x-express-openapi-disable-validation-middleware': true` - Disables input
 validation middleware.
 
+#### Express Middleware to Validate Responses on All Routes
+
+By overriding the implementation of `res.send` within a middleware function, it is possible to perform `res.validateResponse` on all paths, and allow the paths to use `res.send` or `res.json`. The example below is how such a middleware might be added to `app.js`. For a TypeScript example see [TypeScript Example of Global Response Validation](#typescript-example-of-global-response-validation).
+
+```javascript
+function ValidateAllResponses(req, res, next) {
```
jsdevel: can you lowercase this?  typically in javascript constructor functions are PascalCase.

packages/express-openapi/README.md
```diff
@@ -368,6 +368,57 @@ available in the affected operation handler method.
...
+    const strictValidation = req.apiDoc['x-express-openapi-validation-strict'] ? true : false;
+    if (typeof res.validateResponse === 'function') {
+        res.send = WrapSendFunction.bind(res, res.send, res.validateResponse);
```
jsdevel: can you do something like this instead?

```
if (typeof res.validateResponse === 'function') {
  const originalSend = res.send;
  res.send = function expressOpenAPISend(...) // helps to see the function name when debugging

}
```

packages/express-openapi/README.md
```diff
@@ -1051,6 +1102,88 @@ POST.apiDoc = {
 };
 ```
 
+#### TypeScript Example of Global Response Validation
```
jsdevel: i don't feel we need a typescript example.  most users will know how to port it to typescript.

jsdevel:
thanks for doing this!  can you add an example to `koa-openapi`'s README as well?

maxwellb:
Will work on your comments in the next ~48 hours. Thank you for the feedback.

I will have to spend some time reading on koa. I haven't used it yet but IIRC from the docs, express knowledge should be portable.

maxwellb:
I will add new commits, but let me know if you'd like all commits folded in to one/force updated.

packages/express-openapi/README.md
maxwellb: I felt that with the function definition outside of the if-block and assignment of res.send that it might be more readable. Also I wasn't sure if res.json or another function would also need to be wrapped and reuse the logic. Is there a benefit to avoiding the use of bind?

I will update as you suggest. In testing, I didn't need to change res.json. And if someone has another need to wrap a separate function, they can refactor at that time.

packages/express-openapi/README.md
maxwellb: I'm wondering if you would consider keeping it. Here's my logic:

1. Can the same be said for the existing TypeScript example? If anything, the library itself has awesome and extensive type definitions, so it seems like TypeScript is a first-class citizen.
2. If the examples are here to address a potentially common use case, does the benefit of reducing friction outweigh the addition of ~81 lines in a document which is already longer than 1100 lines? Maybe the documentation would be served by a refactor into multiple files. But even if a developer _can_ port it and refactor to fit types/interfaces/etc., should they _have to_?

packages/express-openapi/README.md
```diff
@@ -368,6 +368,54 @@ available in the affected operation handler method.
...
+#### Express Middleware to Validate Responses on All Routes
+
+By overriding the implementation of `res.send` within a middleware function, it is possible to perform `res.validateResponse` on all paths, and allow the paths to use `res.send` or `res.json`. The example below is how such a middleware might be added to `app.js`.
+
+```javascript
+function validateAllResponses(req, res, next) {
+    const strictValidation = req.apiDoc['x-express-openapi-validation-strict'] ? true : false;
+    if (typeof res.validateResponse === 'function') {
+        const send = res.send;
+        res.send = function expressOpenAPISend(...rest) {
```
jsdevel: can the parameter be names `args` instead?

packages/express-openapi/README.md
```diff
@@ -368,6 +368,54 @@ available in the affected operation handler method.
...
+        const onlyWarn = !strictValidation;
```
jsdevel: can you indent the code inside this method?

maxwellb:
> thanks for doing this! can you add an example to `koa-openapi`'s README as well?

How would you feel about merging without this one? #11 is opened against express-openapi, and I would need to build a new koa app to test. Maybe it can be a follow-up PR if it's desirable.

jsdevel:
> How would you feel about merging without this one? #11 is opened against express-openapi, and I would need to build a new koa app to test. Maybe it can be a follow-up PR if it's desirable.

That's fine

jsdevel:
thanks @maxwellb !  this will be helpful

Issue #312, opened 2019-01-15
Properties of BaseSchemaObject


HerrRomich:
According to JSON Schema specification elements of Object "properties" could be either Schema Objects 
 or References. This means for me that in "[open-api/packages/openapi-types/index.ts](https://github.com/kogosoftwarellc/open-api/blob/master/packages/openapi-types/index.ts)" instead of Schema Object:
```typescript
  interface BaseSchemaObject {
    // JSON schema allowed properties, adjusted for OpenAPI
    title?: string;
...
    properties?: {
      [name: string]: SchemaObject;
    };
...  
    example?: any;
    deprecated?: boolean;
  }
```

should be a union:
``` typescript
  interface BaseSchemaObject {
    // JSON schema allowed properties, adjusted for OpenAPI
    title?: string;
...
    properties?: {
      [name: string]: ReferenceObject | SchemaObject;
    };
...  
    example?: any;
    deprecated?: boolean;
  }
```
Thank you

HerrRomich:
Should I have a permission to create a branch?

jsdevel:
of course!

jsdevel:
oh, you can fork this repo and open a PR.

jsdevel:
closed by #313

PR #313, opened 2019-01-15
openapi-types: Allowing to set a property of BaseSchemaObject as a re…

HerrRomich:
…ference to another SchemaObject
Addresses #312


jsdevel:
thanks!

Issue #61, opened 2016-11-05
Console.log ??


memelet:
I notice that all logging is hard-coded to use Console. Is there some way to redirect this modules console output to a application specific logger so it can be routed to eg, a file then ELK, like all the other logs in our api?

jsdevel:
Hmm.  I see what you mean.  Have any ideas?  Maybe a console provider option in the initialize args?

memelet:
Since almost all logging modules support the basic api log.info,warn,error(string) take an logging instance in the initialize(). This would allow the use of bunyan or winston. And even if the logger did not support just that interface it would be trivial to map into it. 

It's too bad that node does not have something like slf4j.

jsdevel:
Yea.  I'd love it if you could submit a PR with support for such an interface in `.initialize(args)` @memelet .

memelet:
If we end up with swagger vs raml, I will do so. Thanks!

mbsimonovic:
I'd like to give this one a crack. Any ideas on how/what it should look like? Guess a TypeScript Logger interface, and a couple of implementations for popular packages such as bunyan, winston, ..

mbsimonovic:
this again spans multiple packages, so i'd like to agree on a plan before making any changes. 

I was thinking something along these lines (please feel free to change my typescript, it's pretty noob);

```
// in express-openapi

export interface GenericLogger { //there's already one called Logger
  error(message: string, ...extra: any[]): void;
  warn(message: string, ...extra: any[]): void;
  info(message: string, ...extra: any[]): void;
  debug(message: string, ...extra: any[]): void;
}

class NoopLogger implements GenericLogger {
  public debug(message: string, ...extra: any[]): void {
    // noop
  }

  public error(message: string, ...extra: any[]): void {
    // noop
  }

  public info(message: string, ...extra: any[]): void {
    // noop
  }

  public warn(message: string, ...extra: any[]): void {
    // noop
  }
}
export interface ExpressOpenAPIArgs extends OpenAPIFrameworkArgs {
  app: Application;
  consumesMiddleware?: { [mimeType: string]: RequestHandler };
  docsPath?: string;
  errorMiddleware?: ErrorRequestHandler;
  exposeApiDocs?: boolean;
  promiseMode?: boolean;
  securityFilter?: RequestHandler;
  logger?: GenericLogger;
}
export function initialize(args: ExpressOpenAPIArgs): OpenAPIFramework {
[...]

PR #322, opened 2019-01-23
openapi-framework: add Logger (#61)

mbsimonovic:


mbsimonovic:
console.debug doesn't exist in node v6... how to you usually handle these?

packages/openapi-framework/index.ts
```diff
@@ -6,6 +6,7 @@ import OpenAPIResponseValidator from 'openapi-response-validator';
 import OpenAPISchemaValidator from 'openapi-schema-validator';
 import OpenAPISecurityHandler from 'openapi-security-handler';
 import { OpenAPI, OpenAPIV2, OpenAPIV3 } from 'openapi-types';
+import { dummyLogger } from 'ts-log';
```
jsdevel: the dummyLogger doesn't actually do anything

packages/openapi-framework/test/constructor.ts
```diff
@@ -181,5 +181,10 @@ describe('OpenapiFramework', () => {
         });
       });
     });
+
+    it('should accept console as a logger', () => {
+      options.logger = console;
+      new OpenapiFramework(options);
```
jsdevel: instead of dummyLogger instead of the default, can we use console?

mbsimonovic: sure

mbsimonovic:
what shall i do with the missing `console.debug` in node v6?

mbsimonovic:
https://github.com/kallaspriit/ts-log/issues/1

packages/openapi-framework/index.ts
```diff
@@ -76,11 +76,21 @@ export default class OpenAPIFramework implements IOpenAPIFramework {
...
     this.featureType = args.featureType;
     this.loggingPrefix = args.name ? `${this.name}: ` : '';
+    this.logger = args.logger ? args.logger : console;
+    // monkey patch for node v6:
+    if (!this.logger.debug) {
+      this.logger.debug = this.logger.log;
+    }
+    this.logger.debug(
+      `${this.loggingPrefix}constructing with`,
+      JSON.stringify(args)
```
jsdevel: not sure how crazy i am about the `JSON.stringify(args)` here.  can you remove it?

packages/openapi-framework/index.ts
```diff
@@ -76,11 +76,21 @@ export default class OpenAPIFramework implements IOpenAPIFramework {
...
   private validateApiDoc;
   private validator;
+  private logger;
 
   constructor(protected args = {} as OpenAPIFrameworkConstructorArgs) {
     this.name = args.name;
     this.featureType = args.featureType;
     this.loggingPrefix = args.name ? `${this.name}: ` : '';
+    this.logger = args.logger ? args.logger : console;
+    // monkey patch for node v6:
```
jsdevel: perhaps we should just remove v6 support?  it's going to be EOL in 3 months anyway.

packages/openapi-framework/index.ts
jsdevel: i'd actually rather not add this line altogether

packages/openapi-framework/index.ts
mbsimonovic: i'd def vote for that :) it's not like this can't be used on node v6, they just have to wrap console and provide .debug()

packages/openapi-framework/index.ts
mbsimonovic: functions no but all other objects would be nicely serialized. it's much easier to catch an error with enough debugging info printed, especially in production where it's not that easy to debug. I've removed it, perhaps there's a better way.

jsdevel:
thanks @mbsimonovic !

Issue #305, opened 2019-01-10
OpenAPIV3: security handler doesn't work as expected


mbsimonovic:
Perhaps this works but my expectations are wrong:

```yaml
openapi: "3.0.0"
servers:
    - url: http://localhost:3000/v1
      description: local server

# globally apply to all operations by default, and turn off for the public ones
security:
    - ApiKeyAuth: []  # no scopes for api-key scheme
paths:
    /search:
        get:
            summary: Returns folder listing by path
            parameters:
                - name: query
                  in: query
                  required: true
                  schema:
                      type: string
            responses:
                '200':
                    description: Success
                    content:
                        application/json:
                            schema:
                                type: object
components:
    securitySchemes:
        ApiKeyAuth:
            type: apiKey
            in: header
            name: X-API-Key

```
and app configured as:
```
const openapiArgs = {
    app,
    apiDoc: v1ApiDoc,
    exposeApiDocs: true,
    paths: path.resolve(__dirname, './api/paths')
    securityHandlers: { ApiKeyAuth: function (req, res, next) {
        // never gets called
        res.status(401).json({message:'unauthorized'});
   } }
};
openapi.initialize(openapiArgs);
```
Then I would expect the ApiKeyAuth to be invoked when /search gets hit with a GET request. This is not the case now. Am I missing something or this hasn't been implemented for v3?

jsdevel:
work was started on this in https://github.com/kogosoftwarellc/open-api/pull/259, but @jwmur hasn't finished it yet.

mbsimonovic:
allright then, will close this issue

PR #323, opened 2019-01-24
Fixes support for args.securityHandlers in OpenApi 3

mendelk:
For context, see #305 and #259 

Based on @jwmur's work in #259.

@jsdevel Let me know if you need me to fix anything in this PR or add any tests etc.

Thanks!


jsdevel:
looks good!  I'll publish later tonight.  Thanks for the implementation @jwmur!

PR #324, opened 2019-01-24
openapi-response-validator: Deep clone schemas before transforming nullable values.

bcrastnopol:
Previously, the response validator would mutate properties in the api doc where nullable: true for openapi v3 docs. For example,  ```{ name: {type: "string", nullable: true}}``` would become ```{ name: {type: ["string", null]}}```, so when the framework tries to validate the doc after populating paths, it would fail validation. Now  that schema is deep copied with lodash.clonedeep as opposed to shallow copied with the spread operator, the api doc doesn't mutate and stays valid after passing through the response validator.

### DIFF
--- a/packages/openapi-response-validator/index.ts
+++ b/packages/openapi-response-validator/index.ts
@@ -222,7 +222,7 @@ function transformOpenAPIV3Definitions(schema) {
   if (typeof schema !== 'object') {
     return schema;
   }
-  const res = { ...schema };
+  const res = JSON.parse(JSON.stringify(schema));
   recursiveTransformOpenAPIV3Definitions(res);
   return res;
 }


packages/openapi-response-validator/index.ts
```diff
@@ -222,7 +223,7 @@ function transformOpenAPIV3Definitions(schema) {
   if (typeof schema !== 'object') {
     return schema;
   }
-  const res = { ...schema };
+  const res = cloneDeep(schema);
```
jsdevel: i know this might be ridiculous, but to avoid adding a dependency, would this suffice?

```
const res = JSON.parse(JSON.stringify(schema));
```

bcrastnopol: I went with the lodash function because the JSON stringify/parse method breaks on non json objects like functions or dates, but since these schemas are probably always going to be valid json I'm fine with the JSON thing.

jsdevel: i imagine it would be a good thing if it failed due to someone passing in a function here.  would you agree?

bcrastnopol: I do! I would want it to be clear to the user why it failed though. JSON.stringify just erases functions silently so it won't be obvious to the user depending on that function why it isn't there. I think the JSON method is fine for this use case though since these things should 
hopefully get caught earlier.

jsdevel:
Thanks @bcrastnopol !

PR #326, opened 2019-01-26
openapi-request-validator: Handle missing or invalid 'Content-Type'

keijokapp:
`OpenAPIRequestValidator.validate(request)` currently fails with `TypeError` if `Content-Type` header is missing or invalid and `requestBody` is defined in schema.

This PR makes it return `415 Unsupported Media Type` on invalid `Content-Type` and `400 Bad Request` if `Content-Type` is missing but required by `requestBody`.


jsdevel:
Thanks!

PR #327, opened 2019-01-30
openapi-types: relax security requirement object types

devoto13:
Specification allows arbitrary names for security schemes,
but current types restrict it to 4 hard-coded values.

See https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#security-requirement-object


jsdevel:
thanks!

PR #328, opened 2019-01-30
express-openapi: Test for OpenApi3 Multipart Form consumes middleware

mendelk:
@jsdevel This is the Readme update you requested, along with all the tests I could think of. If you think the tests are too much (they add a dev dependency for `multer`), I can remove them. Let me know your thoughts!


mendelk:
Did that fix it?

PR #329, opened 2019-01-31
openapi-request-validator: Test for OpenApi3 multipart - request body…

mbsimonovic:
I've removed multer as a dependency and the whole test for multipart from express-openapi because the "bug" can be easily reproduced in the request validator (just uncomment `required: ['files'],`)


mendelk:
@mbsimonovic I don't think the Multer transformation validation should be handled at the level of request validator, as it's an app-level decision. Another user of this library may be using an alternative to Multer etc.

packages/openapi-request-validator/test/data-driven/accept-mutipart-with-boundary.js
```diff
@@ -7,6 +7,7 @@ module.exports = {
         'multipart/form-data': {
           schema: {
             type: 'object',
+            required: ['files'],
```
jsdevel: @mbsimonovic does this reproduce the bug?

packages/express-openapi/package.json
```diff
@@ -33,7 +33,6 @@
   },
   "devDependencies": {
     "@types/ajv": "^1.0.0",
-    "@types/express": "^4.16.0",
-    "multer": "^1.4.1"
```
jsdevel: i don't see any harm in having tests using multer.  it isn't going to effect users in production, and it does provide some checking to ensure that the framework _can work_ with file attachments.

packages/openapi-request-validator/test/data-driven/accept-mutipart-with-boundary.js
mbsimonovic: yes but the request has to be modified as well: `files: ['binary file 1']` will satisfy this required clause so it does validate. i didn't want for the build to fail on this pr, so i've made it pass by setting files to an array of strings. If multer is used, then this won't be a list of strings, but list of objects, so the test will fail just like in the example by @mendelk

jsdevel:
going to close this as i don't see anything wrong with multer as a devDependency and verifying that it works

PR #330, opened 2019-01-31
openapi-framework: replace console with Logger (#61)

mbsimonovic:
follow up on https://github.com/kogosoftwarellc/open-api/pull/322

### DIFF
--- a/packages/openapi-framework/index.ts
+++ b/packages/openapi-framework/index.ts
@@ -8,6 +8,7 @@ import OpenAPISecurityHandler from 'openapi-security-handler';
 import { OpenAPI, OpenAPIV2, OpenAPIV3 } from 'openapi-types';
 import BasePath from './src/BasePath';
 import {
+  ConsoleDebugAdapterLogger,
   IOpenAPIFramework,
   OpenAPIFrameworkAPIContext,
   OpenAPIFrameworkArgs,
@@ -83,10 +84,10 @@ export default class OpenAPIFramework implements IOpenAPIFramework {
     this.name = args.name;
     this.featureType = args.featureType;
     this.loggingPrefix = args.name ? `${this.name}: ` : '';
-    this.logger = args.logger ? args.logger : console;
+    this.logger = args.logger ? args.logger : new ConsoleDebugAdapterLogger();
     // monkey patch for node v6:
     if (!this.logger.debug) {
-      this.logger.debug = this.logger.log;
+      this.logger.debug = this.logger.info;
     }
 
     [
@@ -201,6 +202,7 @@ export default class OpenAPIFramework implements IOpenAPIFramework {
 
     if (this.paths) {
       paths = [].concat(this.paths);
+      this.logger.debug(`${this.loggingPrefix}paths=`, paths);
       paths.forEach(pathItem => {
         if (byString(pathItem)) {
           pathItem = toAbsolutePath(pathItem);
@@ -297,6 +299,7 @@ export default class OpenAPIFramework implements IOpenAPIFramework {
         }
       });
     }
+    this.logger.debug(`${this.loggingPrefix}routes=`, routes);
 
     // Check for duplicate routes
     const dups = routes.filter((v, i, o) => {
@@ -318,6 +321,7 @@ export default class OpenAPIFramework implements IOpenAPIFramework {
 
     routes.forEach(routeItem => {
       const route = routeItem.path;
+      this.logger.debug(`${this.loggingPrefix}setting up`, route);
       const pathModule = injectDependencies(
         routeItem.module.default || routeItem.module,
         this.dependencies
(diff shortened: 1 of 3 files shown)


packages/openapi-framework/index.ts
```diff
@@ -227,6 +228,7 @@ export default class OpenAPIFramework implements IOpenAPIFramework {
       }
     });
     routes = routes.sort(byRoute);
+    this.logger.debug(`${this.loggingPrefix}routes=`, routes.map(r => r.path));
```
jsdevel: i'm not crazy about this because `routes.map` is going to get executed even if debug logging isn't enabled.

packages/openapi-framework/index.ts
```diff
@@ -187,6 +187,7 @@ export default class OpenAPIFramework implements IOpenAPIFramework {
         : null;
 
     const paths = [].concat(this.paths);
+    this.logger.debug(`${this.loggingPrefix}paths=`, paths);
```
jsdevel: with node `console.debug`, is there a way to disable the output in the logs?  the [docs](https://nodejs.org/api/console.html#console_console_debug_data_args) suggest that `console.debug` is merely an alias for `console.log`

mbsimonovic:
but this is only once when the framework is initialized

jsdevel:
> but this is only once when the framework is initialized

while that is true, we're still adding log output for debug purposes whether or not the user wishes to actually debug this framework.  i'd rather see us use https://www.npmjs.com/package/debug if our goal is to make it easier to debug the framework.  debug logs should be opt in.

mbsimonovic:
i completely agree that debug logs should be opt in, but instead of going  https://www.npmjs.com/package/debug <https://www.npmjs.com/package/debug>, how about we don’t assign console by default, but just use the dummy no-op logger?

mbsimonovic:
btw, the debug package doesn't fit the Logger interface, so I would still prefer logger.debug messages, and pass my own logger (winston/bunyan), which outputs everything in json, and i collect all my logs for central processing. i would actually keep the level on debug even in production, but then discard old debug logs after a week or so. It's better to have more than not enough info on what's going on. The performance hit is close to zero, these logs are async anyways. 

So how about we define a custom logger, that adapts `console` so it doesn't do anything for .debug() ?

jsdevel:
> So how about we define a custom logger, that adapts `console` so it doesn't do anything for .debug() ?

not sure, but i'm open to suggestions.  i just want to make sure that debug logs are opt-in, not opt-out, and that we don't add side effects that have a tremendous amount of performance overhead i.e.

```
console.debug('something', paths.map(/*   */)); // .map is called whether debug logs are enabled or not.
```

mbsimonovic:
@jsdevel what do you think now?

packages/openapi-framework/src/types.ts
```diff
@@ -15,6 +15,34 @@ export {
...
+   * This class delegates to `console` and overrides `.debug()` to be a no-op.
+   * @param message
+   * @param optionalParams
+   */
+  public debug(message?: any, ...optionalParams: any[]): void {
+    // no-op
+  }
+
+  public error(message?: any, ...optionalParams: any[]): void {
+    console.error(message, optionalParams);
```
jsdevel: i think we need to spread the optionalParams

mbsimonovic: what do you mean, `console.error(message, ...optionalParams);` ?

jsdevel: > what do you mean, `console.error(message, ...optionalParams);` ?

yes.  otherwise we're passing an array to the console methods correct?

mbsimonovic: that's correct, now looks better

packages/openapi-framework/src/types.ts
```diff
@@ -15,6 +15,34 @@ export {
   OpenAPIErrorTransformer
 };
 
+export class ConsoleDebugAdapterLogger implements Logger {
+  /**
+   * `console.debug` is just an alias for `.log()`, and we want debug logging to be optional.
+   * This class delegates to `console` and overrides `.debug()` to be a no-op.
+   * @param message
```
jsdevel: can you remove these 2 `@param` comments?

jsdevel:
@mbsimonovic looks good!  left a minor comment, and need to see the conflict resolved.  i'll be ready to merge once that's complete

packages/openapi-framework/src/types.ts
mbsimonovic: sure

mbsimonovic:
> @mbsimonovic looks good! left a minor comment, and need to see the conflict resolved. i'll be ready to merge once that's complete

i've rebased onto master and resolved the conflicts, you should be able to merge w/o any issues.

mbsimonovic:
@jsdevel the build failed in `npm run coveralls` for node v6, my commits have got nothing to do with this right:

```
/home/travis/build/kogosoftwarellc/open-api/node_modules/coveralls/bin/coveralls.js:18
        throw err;
        ^
Bad response: 422 {"message":"Couldn't find a repository matching this job.","error":true}

```

jsdevel:
@mbsimonovic every now and then coveralls isn't reachable.  i usually just re-run the job that filed and it passes.  that's what i did in this case.  thanks for your patience and persistence!

jsdevel:
published

PR #331, opened 2019-01-31
express-openapi: Example showing multer only when needed

mendelk:
Plus minor bug fixes.

This is to address some of @mbsimonovic 's points in #279 ([relevant comment](https://github.com/kogosoftwarellc/open-api/issues/279#issuecomment-459291043) and onwards).

As the implementation of `args.consumesMiddleware` get's more and more complex to handle files, I'm leaning towards this being handled by `express-openapi` automatically for file uploads.

Thoughts @mbsimonovic and @jsdevel ?


packages/express-openapi/index.ts
```diff
@@ -116,8 +116,6 @@ export function initialize(args: ExpressOpenAPIArgs): OpenAPIFramework {
       let middleware = [].concat(ctx.additionalFeatures);
 
       if (operationDoc && ctx.allowsFeatures) {
-        middleware.unshift(createAssignApiDocMiddleware(apiDoc, operationDoc));
```
mendelk: I moved this way up so that `consumesMiddleware` has access to `req.operationDoc`

jsdevel: before we were checking to ensure that `operationDoc` wasn't falsey.

mbsimonovic:
i would love to be able to add middleware on particular paths, like with `x-express-openapi-additional-middleware`, but specify if it should run before or after validation.

mendelk:
@mbsimonovic Maybe there can be introduced the concept of "middleware hooks" such as `pre-request-validation`, `pre-response-validation` etc, scoped like `x-express-openapi-additional-middleware`

mendelk:
@jsdevel If I have time over the weekend, I'll take a crack at it.

packages/express-openapi/test/sample-projects/with-consumes-middleware-multipart-openapi3/spec.js
```diff
@@ -10,14 +10,14 @@ describe('with openapi3 multipart form consumes middleware', function() {
   it('should work with single attachment', function(done) {
     request(app)
       .post('/v3/single-attachment')
-      .attach('file', new Buffer('Some test text', 'utf-8'), 'file_a.txt')
```
jsdevel: @mendelk is there a way to avoid modifying existing tests?

jsdevel:
@mendelk i'd like to merge this.  any chance you can respond to the feedback?

jsdevel:
going to close this due to inactivity.  @mendelk if you have time to re-open and address the feedback feel free.  thanks for your support!

PR #347, opened 2019-02-11
Property items can't be array

MrFranke:
> items - Value MUST be an object and not an array. Inline or referenced schema MUST be of a Schema Object and not a standard JSON Schema. items MUST be present if the type is array.

Check openapi 3 spec: https://swagger.io/specification/


jsdevel:
@MrFranke the IJsonSchema type represents [JsonSchema items](https://json-schema.org/latest/json-schema-validation.html#rfc.section.6.4.1), not swagger/openapi.  The change as you've proposed would break a lot of things

Issue #352, opened 2019-02-12
args.operations does not handle multiple operations per route


sebasmagri:
When a path has multiple operations defined, openapi-framework is going to produce duplicate routes.

jsdevel:
@sebasmagri can this be closed?

sebasmagri:
@jsdevel yes, fixed on #353

PR #353, opened 2019-02-12
feat: different approach to handle operations in openapi-framework initialization

sebasmagri:
This initially implements a different approach for `args.operations` handling so that it's easier to handle multiple methods per path, which is not fully working as per #352.

Current issue is mainly with tests `operations-dir-with-overlap` and `operations-dir`, which have overlapping/duplicated routes due to `paths` and `operations` being used at the same time. What should be the behavior for those cases?

### DIFF
--- a/packages/openapi-framework/index.ts
+++ b/packages/openapi-framework/index.ts
@@ -244,31 +244,36 @@ export default class OpenAPIFramework implements IOpenAPIFramework {
     if (this.operations) {
       const apiDocPaths = this.apiDoc.paths;
       Object.keys(apiDocPaths).forEach(apiDocPathUrl => {
-        const route = routes.filter(v => v.path === apiDocPathUrl)[0];
         const pathDoc = apiDocPaths[apiDocPathUrl];
-        Object.keys(pathDoc)
-          .filter(byMethods)
-          .forEach(method => {
-            const methodDoc = pathDoc[METHOD_ALIASES[method]];
-            const operationId = methodDoc.operationId;
-            if (operationId && operationId in this.operations) {
-              const operation = this.operations[operationId];
-              if (route) {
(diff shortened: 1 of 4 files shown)


packages/openapi-framework/index.ts
```diff
@@ -244,31 +244,36 @@ export default class OpenAPIFramework implements IOpenAPIFramework {
...
+          path: apiDocPathUrl,
+          module: Object.keys(pathDoc)
+            .filter(byMethods)
+            .reduce((acc, method) => {
+              const methodDoc = pathDoc[METHOD_ALIASES[method]];
+              const operationId = methodDoc.operationId;
+              if (operationId && operationId in this.operations) {
+                const operation = this.operations[operationId];
+
+                console.log(operation);
```
jsdevel: please remove the console logs

packages/openapi-framework/index.ts
```diff
@@ -244,31 +244,36 @@ export default class OpenAPIFramework implements IOpenAPIFramework {
...
+
+                acc[METHOD_ALIASES[method]] = (() => {
+                  let _f: any = operation;
```
jsdevel: please remove the `_` prefix.

jsdevel:
@sebasmagri the overall approach looks good to me.  the build is broken

sebasmagri:
@jsdevel right, I'd like to know what the approach would be when we have both `paths` and `operations`. What do you think it should be?

jsdevel:
> What do you think it should be?

I'd say operations would take precedence over paths with a warning.  Just so long as we don't break back compat.

sebasmagri:
@jsdevel can you give this a check up again?

jsdevel:
Looks good!  Thanks for the support!

PR #357, opened 2019-02-13
Dependency injection for operations

sebasmagri:
Inject dependencies on operations by the means of binding the operation handlers so that they get `this.dependencies`

### DIFF
--- a/packages/express-openapi/README.md
+++ b/packages/express-openapi/README.md
@@ -742,6 +742,33 @@ initialize({
 app.listen(3000);
 ```
 
+Operations also get `args.dependencies` injected as
+`this.dependencies` on the function scope.
+
+```js
+// ./app.js
+import express from 'express';
+import { initialize } from 'express-openapi';
+
+const app = express();
+
+initialize({
+  app,
+  apiDoc: './apiDoc.yml',
+  dependencies: {
+    log: console.log
+  },
+  operations: {
+    getFoo: function(req, res) {
+      this.log('calling request handler');
+      res.send('foo');
+    }
+  }
+});
+
+app.listen(3000);
+```
+
 **Note:** Handlers in args.paths take precedence over handlers in args.operations for
 historical reasons.
 
--- a/packages/express-openapi/README.md
+++ b/packages/express-openapi/README.md
@@ -766,7 +766,7 @@ initialize({
   },
   operations: {
     getFoo: function(req, res) {
-      this.log('calling request handler');
+      this.dependencies.log('calling request handler');
       res.send('foo');
     }
   }
@@ -774,15 +774,6 @@ initialize({
 
 app.listen(3000);
 ```
-
-**Note:** `args.dependencies` is added to the operation's `this`
-context via `operation.bind({dependencies: args.dependencies})`.  This
-requires your operations to be regular [function
-expressions](https://developer.mozilla.org/en-US/docs/web/JavaScript/Refeerence/Operators/function)
-and not [arrow
-functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions)
-(due to the fact that `this` is lexical to the surrounding scope in an
-arrow function and cannot be bound to anything else). 
 
 #### args.pathSecurity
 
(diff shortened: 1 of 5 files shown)


sebasmagri:
@jsdevel sure, was adding some notes but wanted some feedback in advance...

packages/openapi-framework/index.ts
```diff
@@ -260,7 +260,10 @@ export default class OpenAPIFramework implements IOpenAPIFramework {
                 acc[METHOD_ALIASES[method]] = (() => {
                   const innerFunction: any = operation;
                   innerFunction.apiDoc = methodDoc;
-                  return innerFunction;
+                  // Operations get dependencies injected in `this`
+                  return innerFunction.bind({
+                    dependencies: this.dependencies
```
jsdevel: also, perhaps we should create a new object, so the operations can't modify the reference?

i.e.

```
  dependencies: {...this.dependencies}
```

packages/express-openapi/README.md
```diff
@@ -743,7 +743,8 @@ app.listen(3000);
 ```
 
 Operations also get `args.dependencies` injected as
-`this.dependencies` on the function scope.
+`this.dependencies` on the function scope. This requires the operation
```
jsdevel: can you reword this to be:

```
**Note:** `args.dependencies` is added to the operation's `this` context via `operation.bind({dependencies: args.dependencies})`.  This requires your operations to be regular [function expressions](https://developer.mozilla.org/en-US/docs/web/JavaScript/Reference/Operators/function) and not [arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) (due to the fact that `this` is lexical to the surrounding scope in an arrow function and cannot be bound to anything else).
```

sebasmagri: Sure!

packages/express-openapi/README.md
```diff
@@ -742,8 +742,42 @@ initialize({
 app.listen(3000);
 ```
 
-**Note:** Handlers in args.paths take precedence over handlers in args.operations for
-historical reasons.
+Operations also get `args.dependencies` injected as
```
jsdevel: Same here

packages/express-openapi/README.md
```diff
@@ -742,8 +742,47 @@ initialize({
...
+      this.log('calling request handler');
+      res.send('foo');
+    }
+  }
+});
+
+app.listen(3000);
+```
+
+**Note:** `args.dependencies` is added to the operation's `this`
```
jsdevel: Sorry, i thought this section was for args.paths.  You can remove this Note, and the one in args.paths if the one in args.paths is no longer valid.

sebasmagri:
@jsdevel Hi! I've applied the requested changes in docs. Anything else you think we should consider?

packages/express-openapi/README.md
```diff
@@ -742,8 +742,47 @@ initialize({
...
+
+initialize({
+  app,
+  apiDoc: './apiDoc.yml',
+  dependencies: {
+    log: console.log
+  },
+  operations: {
+    getFoo: function(req, res) {
+      this.log('calling request handler');
```
jsdevel: should this be `this.dependencies.log`?  If not, then the bind example in the note above should be `.bind({...args.dependencies`

sebasmagri:
> sorry to have so much back and forth over this. please look at the file as a whole and determine what makes sense. i'm eager to merge this!

You're right! I'm gotta check it out.

sebasmagri:
@jsdevel I think it's much better now.

jsdevel:
thanks @sebasmagri !  is the note about operations taking precedence in args.paths still valid?

sebasmagri:
@jsdevel it was not anymore as per [comment](https://github.com/kogosoftwarellc/open-api/pull/353#issuecomment-462850271). I updated the readme to reflect it.

jsdevel:
You killed it @sebasmagri!  Thanks!

sebasmagri:
@jsdevel great! thanks a lot... looking forward to have this in a release ;-)

jsdevel:
published in the latest version @sebasmagri

sebasmagri:
Thanks @jsdevel

PR #363, opened 2019-02-20
fs-routes: add logger(#61)

mbsimonovic:
i had to add ts-log as a dependency, doesn't work as a peerDependency, because the exported interface changes


jsdevel:
sorry, but we're not using logging here, and the addition of debug logs here isn't something i'd like to add.  the ability to change the logger wasn't meant for debugging these packages.  i'd be open to adding the https://www.npmjs.com/package/debug package instead.

PR #364, opened 2019-02-20
openapi-request-validator: add logger(#61)

mbsimonovic:


packages/openapi-request-validator/index.ts
```diff
@@ -182,6 +191,7 @@ export default class OpenAPIRequestValidator
     let err;
     let schemaError;
     let mediaTypeError;
+    this.logger.debug('validating', request);
```
jsdevel: please use https://www.npmjs.com/package/debug for debug log messages and remove these.

packages/openapi-request-validator/index.ts
```diff
@@ -314,12 +357,18 @@ function extendedErrorMapper(mapper) {
 
 function getSchemaForMediaType(
   contentTypeHeader: string,
-  requestBodySpec: OpenAPIV3.RequestBodyObject
+  requestBodySpec: OpenAPIV3.RequestBodyObject,
+  logger: Logger
 ): string {
+  if (!contentTypeHeader) {
```
jsdevel: why are we adding this without a test?

packages/openapi-request-validator/index.ts
mbsimonovic: i'm not using them for debugging, but to track how my app behaves in production, so all the logs (json, produced by a bunyan), from more than one app, are collected and pushed to a central logging server. the debug package wouldn't work for me in this case. I could change these to .info, but we said logging should be opt-in.

packages/openapi-request-validator/index.ts
mbsimonovic: Test if a logger gets passed? How can anything fail there? Or you're asking about `if (!contentTypeHeader)`? That's just refactoring, I've moved the conditional from `validate` because it makes more sense to check it here (so that every possible caller of getSchemaForMediaType doesnt have to perform this check), and it makes the code look better:
```
      const mediaTypeMatch =
        contentType && getSchemaForMediaType(contentType, this.requestBody);
``` 
becomes:
```
      const mediaTypeMatch = getSchemaForMediaType(contentType, this.requestBody);
```

jsdevel:
@mbsimonovic if there's a way to _only_ pass the logger at this time, without adding or altering existing log messages, then i'd love to see progress on this.

the addition of debug logs needs to be taken into consideration with performance logging too, which will most likely require flags to be passed in via args.  we can have that discussion at a later time though.

i appreciate your support so far, and hope to close #61 soon.

mbsimonovic:
i bet performance hit is not an issue, but i'm open to be proven wrong :) it doesn't seem we'll see eye to eye on this and i'd like to move on so i'll remove extra .debug calls

packages/openapi-request-validator/index.ts
```diff
@@ -321,12 +331,17 @@ function extendedErrorMapper(mapper) {
...
+    return;
+  }
   let contentType: string;
   try {
     contentType = contentTypeParser.parse(contentTypeHeader).type;
   } catch (e) {
+    logger.warn('failed to parse content-type', contentTypeHeader, e);
```
jsdevel: can you add the loggingKey here?

packages/openapi-request-validator/index.ts
jsdevel: sorry for the late reply, i mean the `if` condition.  can you add a test that covers that?

packages/openapi-request-validator/index.ts
```diff
@@ -53,6 +56,10 @@ export default class OpenAPIRequestValidator
       throw new Error(`${loggingKey}missing args argument`);
     }
 
+    if (args.logger) {
```
jsdevel: can you add a test that passes a logger?

jsdevel:
@mbsimonovic i'd love to get this merged.

mbsimonovic:
@jsdevel sorry i've been super busy, will get back to this by next monday latest

mbsimonovic:
loggingKey and a test added, any other suggestions?

packages/openapi-request-validator/test/tests.ts
```diff
@@ -9,6 +9,11 @@ describe(require('../package.json').name, () => {
     new Sut(initialArguments);
     expect(initialArguments).to.eql(createArguments());
   });
+  it('should accept a logger', () => {
+    /* tslint:disable:no-unused-expression */
+    // @ts-ignore
+    new Sut({ logger: console, schemas: [{}] });
```
jsdevel: this test covers the constructor, but changes were made to the `#validate` method.  can you add a test that covers those changes?

mbsimonovic: the only change apart from logging statements is in `getSchemaForMediaType()`:
```
  if (!contentTypeHeader) {
    return;
  }
```

and there's a test that already exercises this conditional (should handle missing content type as missing body)

jsdevel: ok.  thanks!

Issue #371, opened 2019-02-27
Invalid number parameters pass validation


Manny651:
When a query or path parameter is defined as a number but the actual value cannot be converted (e.g. 'abc'), the parameter is converted to NaN and passes the request validator. NaN shouldn't be a valid value for a number (or integer) type.

jsdevel:
indeed!  please submit a PR to fix this issue.

Manny651:
I'm not sure how this case should be handled: should the request coercer convert the invalid value to null, or throw an error, or is it a request validator issue?

jsdevel:
that's a great question.  i think i lean towards coercer.  how about you @Manny651 ?

jsdevel:
i wonder though, would there be cases where you'd want to respond with a validation error?

jsdevel:
perhaps the coercer shouldn't override the value if it can't be coerced into it's intended type?

Manny651:
I think the best solution would be to prevent the coercer from overriding values that cannot be converted. That way, the request validator would respond with the appropriate error message for that case.

jsdevel:
> I think the best solution would be to prevent the coercer from overriding values that cannot be converted.

i think i agree with that.  otherwise, the validator might be making assumptions and wouldn't have all the info to respond with an informative error.

please go ahead and update the coercer then.

jsdevel:
published

PR #373, opened 2019-02-28
Don't coerce into numbers values that cannot be converted (issue #371)

Manny651:
Possible fix for #371


jsdevel:
Looks like lint errors

jsdevel:
thanks @Manny651 !  i'll publish later tonight

PR #377, opened 2019-03-04
express-openapi: Set default basePath (issue #376)

### LINKED ISSUE #376: Generated API doc fails Swagger validation when basePath is not set
When configuring express-openapi, if the API's basePath option is not set, the generated API doc has the basePath property set to an empty string, which is invalid according to the swagger.io's online validator (the one used by default in Swagger UI).

According to the spec https://swagger.io/docs/specification/2-0/api-host-and-base-path/ : 
> [basePath] must start with a leading slash /. If basePath is not specified, it defaults to /

Manny651:
Fixes #376

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ sudo: false
 node_js:
   - 10
   - 8
-  - 6
 notifications:
   email:
     on_success: "never"
--- a/packages/express-openapi/index.ts
+++ b/packages/express-openapi/index.ts
@@ -95,8 +95,10 @@ export function initialize(args: ExpressOpenAPIArgs): OpenAPIFramework {
             if (req.apiDoc.swagger) {
               // @ts-ignore
               req.apiDoc.host = req.headers.host;
+              const apiBasePath = req.baseUrl + basePath;
               // @ts-ignore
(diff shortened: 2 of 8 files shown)


jsdevel:
can you add this for `koa-openapi` too?

Manny651:
I'm having trouble running the koa-openapi tests; the 8 basic-usage tests pass, then the process hangs forever.

jsdevel:
@Manny651 looks like the `travis-test` script is actually disabled.  not affecting the travis build at all, but prevents me from seeing any error there.

jsdevel:
@Manny651 can you change the `travis-test` script in `koa-openapi/package.json` to `npm run cover`?

Manny651:
It seems that the koa-openapi tests fail on node v6 because they are using async functions (requires node 7)

jsdevel:
@Manny651 looks like the v6 tests are failing on a few specs.  v6 is eol in april.  can you remove v6 from the .travis.yml file?  thanks for going the extra mile!

jsdevel:
thanks @Manny651 !

jsdevel:
published

PR #390, opened 2019-03-21
Accept a requestBody with missing readOnly required property in ref (fixes #389)

### LINKED ISSUE #389: request validator: readonly & required properties throwing validation error in request body
For OpenAPI 3, when defining a schema in the component section with a property that is both readonly and required, the request validator will return a bad request when that property is missing from the request body.

I'm currently handling this by removing all readonly property keys from the required array of the schema before passing it to ajv.compile as the body schema, but it's not very elegant, fragile and untested.

example schema stub for failing validation:

```
    requestBody:
        required: true
        content:
            application/json:
                schema:
                    $ref: '#/components/schemas/Thing'

    schemas:
        Thing:
            required:
                - id
                - name
             properties:
                 id:
                     type: number
                     readonly: true
                 name:
                     type: string
```

The OpenAPI 3 spec allows for this sort of behavior, so I'm unsure whether this is intended or not.

BrandonFitzgibbon:
Fix for when a requestBody is missing a required and readOnly parameter in a referenced schema. Only affects referenced schemas and OpenAPI3.

### DIFF
--- a/packages/openapi-request-validator/index.ts
+++ b/packages/openapi-request-validator/index.ts
@@ -163,9 +163,27 @@ export default class OpenAPIRequestValidator
     if (args.requestBody) {
       /* tslint:disable-next-line:forin */
       for (const mediaTypeKey in args.requestBody.content) {
+        let bodyContentSchema = args.requestBody.content[mediaTypeKey].schema;
+        if ('$ref' in bodyContentSchema) {
+          const objectSchema = v.getSchema(bodyContentSchema.$ref);
+          if (
+            objectSchema &&
+            objectSchema.schema &&
+            objectSchema.schema.properties
+          ) {
+            Object.keys(objectSchema.schema.properties).forEach(prop => {
+              const propertyValue = objectSchema.schema.properties[prop];
+              if (propertyValue.readOnly) {
+                const index = objectSchema.schema.required.indexOf(prop);
+                objectSchema.schema.required.splice(index, 1);
+              }
+            });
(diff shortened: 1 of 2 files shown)


BrandonFitzgibbon:
i'll come back with fixes for the linter

jsdevel:
can you use the `bin/commit` script to create the commit, or add the package name as a prefix to the commit?  i.e. `openapi-request-validator: some message`.  Thanks for adding `fixes #<issue number>` to the commit message!

BrandonFitzgibbon:
Should have read the contribution guidelines. @jsdevel let me know if anything else is needed.

BrandonFitzgibbon:
@jsdevel this needs more work before it can be merged. There's a missing test to reject request bodies based on the $ref schema type. My branch allows an array of objects to bypass validation when the spec asks for an object.

jsdevel:
hmm.  ok, i'll hold off.

BrandonFitzgibbon:
Took a closer look at the problem is with my spec. I didn't explicitly type the referenced schema as an object and so the validator accepted either type. When explicitly typed it rejects as expected.

So my bad.

jsdevel:
ok, so am i good to merge?

BrandonFitzgibbon:
Yep.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

jsdevel:
thanks @BrandonFitzgibbon !

BrandonFitzgibbon:
Thank you for maintaining!

PR #392, opened 2019-03-22
adding (failing) test for response validator when some fields required and some are not

troymolsberry:
This test fails for me. The expectation is that it should permit an object to be returned with only a single attribute named "name". Please let me know if this expectation is not correct. Cheers,


jsdevel:
@troymolsberry not sure.  if you can research the spec and provide a fix that'd be awesome!

jsdevel:
@troymolsberry ?

PR #400, opened 2019-03-27
openapi-request-validator: accept reqBody RO nested ref (fixes #394)

### LINKED ISSUE #394: requestBody fails when missing readonly required properties in nested ref
Related to #389 and opened for visibility.

PR #390 fixed a single depth issue where the requestBody was a single ref and the ref contained the entire schema definition.

I'm working on a PR now that seeks to address the following cases:

- requestBody contains an allOf with a ref
- the refs themselves contain refs to other schemas with required readOnly props
- the requestBody is an array

BrandonFitzgibbon:
Some insight:

- Allows required properties to be missing from requestBody if readOnly
- Works with nested ref
- Works with array where items are ref
- Works with AllOf, AnyOf and OneOf
- Adds tests

I've achieved this within the request-validator package by cloning the schemas, resolving refs and removing all instances of readOnly properties from the required array of their respective schemas.

Edit: Clarity

### DIFF
--- a/packages/openapi-request-validator/index.ts
+++ b/packages/openapi-request-validator/index.ts
@@ -163,27 +163,12 @@ export default class OpenAPIRequestValidator
     if (args.requestBody) {
       /* tslint:disable-next-line:forin */
       for (const mediaTypeKey in args.requestBody.content) {
-        let bodyContentSchema = args.requestBody.content[mediaTypeKey].schema;
-        if ('$ref' in bodyContentSchema) {
-          const objectSchema = v.getSchema(bodyContentSchema.$ref);
-          if (
-            objectSchema &&
-            objectSchema.schema &&
-            objectSchema.schema.properties
-          ) {
-            Object.keys(objectSchema.schema.properties).forEach(prop => {
-              const propertyValue = objectSchema.schema.properties[prop];
-              if (propertyValue.readOnly) {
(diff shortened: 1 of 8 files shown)


jsdevel:
is this allowed in the schema?  can you provide a link to the openapi v3 docs?

BrandonFitzgibbon:
for v3

"You can use the readOnly and writeOnly keywords to mark specific properties as read-only or write-only. This is useful, for example, when GET returns more properties than used in POST – you can use the same schema in both GET and POST and mark the extra properties as readOnly. readOnly properties are included in responses but not in requests, and writeOnly properties may be sent in requests but not in responses."

https://swagger.io/docs/specification/data-models/data-types/

BrandonFitzgibbon:
Also pretty relevant:

"If a readOnly or writeOnly property is included in the required list, required affects just the relevant scope – responses only or requests only. That is, read-only required properties apply to responses only, and write-only required properties – to requests only."

jsdevel:
got it!  once the build passes, i'll merge

BrandonFitzgibbon:
@jsdevel build passed a while back. Anything else you want to take a look at?

jsdevel:
nope!  thanks @BrandonFitzgibbon !  travis was just taking a while to update the status herein and i forgot.  thanks for your support and the reminder!

PR #409, opened 2019-04-03
openapi-request-validator: Set "useDefaults: true" on Ajv

### LINKED ISSUE #404: [openapi-default-setter] Populate defaults in request body
Breaking out into a separate issue from https://github.com/kogosoftwarellc/open-api/pull/398#issuecomment-478292795

The openapi-default-setter package does not support setting defaults in the request body. Ajv (which is used by openapi-request-validator) has a `useDefaults` option, but it sounds like adding support for setting body defaults in the openapi-default-setter package is preferred.

Thoughts on where to begin or prior art to take inspiration from?

TuckerWhitehouse:
Fixes #404

### DIFF
--- a/packages/openapi-request-validator/index.ts
+++ b/packages/openapi-request-validator/index.ts
@@ -83,6 +83,7 @@ export default class OpenAPIRequestValidator
     }
 
     const v = new Ajv({
+      useDefaults: true,
       allErrors: true,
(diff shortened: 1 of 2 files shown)


jsdevel:
Thanks @TuckerWhitehouse !  can you also add this to the response validator in a separate PR?  do we need this there?

PR #412, opened 2019-04-08
Fix broken link in express-openapi/README.md

sidhuko:
Supported Versions of Node was linking to incorrect anchor in express-openapi/README.md


jsdevel:
thanks @sidhuko !

PR #419, opened 2019-04-09
Fix media type checking for request bodies

### LINKED ISSUE #418: Curious file upload issue
May be related to #29, but I'm not sure.

I am uploading files via PUT requests. The following spec works fine for me (ignoring responses):

```json
  doc.put.apiDoc =
  {
    description: 'Asset upload.',
    operationId: 'assetUpload',
    requestBody: {
      description: 'The binary asset data.',
      content: {
        '*/*': {
          schema: {
            type: 'string',
            format: 'binary',
          },
        },
      },
    },
};
```

As expected, if I limit the content type to e.g. `image/*` and try to upload a different file, that is refused. But hang on...

1. If I upload any file and set an unacceptable `Content-Type` header, the upload is rejected. So far, so good.
1. If I upload any file and don't set any `Content-Type` header, the upload is accepted. That seems weird to me.
1. With an acceptable `Content-Type` header If the header does match, I get a strange validation error: `"errorCode":"type.openapi.validation","message":"should be string","location":"body"`.

Now... I know for a fact that I'm just uploading binary data as the spec requests, because if I skip the `Content-Type` header entirely, as I said, everything works.

It's just weird that well-behaving clients seem to be punished, whereas badly behaving clients work.

jfinkhaeuser:
Fixes #418 

The reasoning behind the patch is the following (may be wrong, let me know):

1. If a request body is required, we can error out immediately if there isn't any - there is no need for checking the media type. In fact, checking the media type only if a request body exists seems to contradict the specs.
1. An *empty* request body may indicate that it hasn't been parsed, perhaps because it wasn't to parse. So we need to ensure the request is at least `readable` before accepting it.

The latter point is a little tricky.

From what I can tell, any parsed request body would go into the first branch, where `request.body` is set and contains values. That can then be validated against the schema, and shouldn't break anything.

If `request.body` isn't set - well, honestly I don't know why. So I'm assuming the request body just hasn't been parsed, and the `{}` it's set to is a default value, and this assumption may be wrong.

But based on this assumption, for supporting file uploads that are binary strings, we need a readable request. If that doesn't exist, there must be a mistake.

What this change *does not* do is check that in the latter case, the schema is actually for binary strings - that should probably be added, I guess.

WDYT?


jsdevel:
please add a test and verify the build passes

jsdevel:
also, please use the `bin/commit` script, and ensure you reference the issue in the commit message (as you've done here).  also, please be more descriptive in your commit message.  i rely on the commit messages to generate the CHANGELOG

jfinkhaeuser:
Alright, I'll close this PR and open a new one.

jsdevel:
@jfinkhaeuser better to keep the PR open and use git rebase

PR #430, opened 2019-04-19
express-openapi: Fixes #429, reordering createAssignApiDocMiddleware

### LINKED ISSUE #429: Access to req.operationDoc in securityHandler
I'd like to be able to access req.operationDoc in a securityHandler. I don't need any body parsing or parameter validation at that point, but it would be great to access the operation definition when figuring permissions.

Basically, I want to define a vendor-prefixed schema extension for operation-level permissions, and then access that definition in the central security handler and be able to send a 401 before actually getting to the operation handler.

It looks to me like the result of #286 is that the security middleware is now run before any API doc validation at all, which would make this impossible?

ashtarcommunications:


jsdevel:
Can you update koa too?

ashtarcommunications:
No problem, added that as well.

jsdevel:
Thanks!

PR #466, opened 2019-05-31
openapi-framework: Log an improved warning if the operationId of an operation is not specified (fixes #465)

### LINKED ISSUE #465: Bad error message if operationId is not specified
When calling `initialize` with an `apiDoc` that has an operation that is missing an `operationId`, express-openapi emits a warning and then throws an exception. The warning is:

> Operation undefined not found in the operations parameter

The exception is:

> TypeError: Cannot read property 'undefined' of undefined

Neither is helpful in identifying the problem.

I have traced this issue into openapi-framework, so it is not specific to express-openapi.

michaellperry:
One new test. I opted to change just the warning and not the exception. If you feel that the exception should be changed as well, then an additional guard clause will need to be inserted.

### DIFF
--- a/packages/openapi-framework/index.ts
+++ b/packages/openapi-framework/index.ts
@@ -268,6 +268,12 @@ export default class OpenAPIFramework implements IOpenAPIFramework {
                     dependencies: { ...this.dependencies }
                   });
                 })();
+              } else if (operationId === undefined) {
+                this.logger.warn(
+                  `${
+                    this.loggingPrefix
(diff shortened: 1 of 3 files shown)


jsdevel:
this is great @michaellperry !  thanks!  i'll merge once the build passes.

jsdevel:
@michaellperry can you fix the build?  i'll merge once it passes.

michaellperry:
That did it. Looks like I forgot to run the linter.

jsdevel:
thanks!

PR #471, opened 2019-06-12
openapi-framework: Update customFormat argument typing

mgrunberg:
Extracted from [docs](https://github.com/kogosoftwarellc/open-api/tree/master/packages/express-openapi#argscustomformats):

```
customFormats argument is an object of custom formats. 
Each key is the name of the format to be used with the format
```

With the current type definition, it's an object with only one property name `string`.


jsdevel:
@mgrunberg can you fix the lint error?

mgrunberg:
@jsdevel sorry about that. I forgot to run the linter before making the PR

PR #472, opened 2019-06-14
openapi-request-validator: accept reqBody readOnly in nested properties

littldr:
Hi,

#400 fixed the nested readOnly behaviour for all schemas with refs. But if you are not using refs for nested properties, it will not allow you to use readOnly properties correctly.

This change will call the sanitize function and itself (recursive) for all properties inside the current schema. The recursive call will ensure that all nested properties are resolved and sanitized, no matter how deep they are nested.

Thanks for maintaining this project! ❤️

### DIFF
--- a/packages/openapi-request-validator/index.ts
+++ b/packages/openapi-request-validator/index.ts
@@ -487,7 +487,15 @@ function resolveAndSanitizeRequestBodySchema(
 ) {
   let resolved;
   let copied;
-  if ('$ref' in requestBodySchema) {
+
+  if ('properties' in requestBodySchema) {
+    const schema = requestBodySchema as OpenAPIV3.NonArraySchemaObject;
+    Object.keys(schema.properties).forEach(property => {
+      let prop = schema.properties[property];
+      prop = sanitizeReadonlyPropertiesFromRequired(prop);
+      prop = resolveAndSanitizeRequestBodySchema(prop, v);
+    });
+  } else if ('$ref' in requestBodySchema) {
     resolved = v.getSchema(requestBodySchema.$ref);
     if (resolved && resolved.schema) {
       copied = JSON.parse(JSON.stringify(resolved.schema));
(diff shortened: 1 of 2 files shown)


jsdevel:
@littldr i wish i could leave this pr open as an example for all to see, but sadly i must merge it.  thank you for following the contributor guidelines!

littldr:
Any chance that this will be released soon? Or do you have some kind of timetable when releases happen?

jsdevel:
i'll probably release today.  it usually just depends on demand and how many prs are waiting

AlexHolly:
I get Range/Stack Error in version "openapi-request-validator": "3.8.2".  I think it is related to this PR, do I have to change something or is it a bug? It works in "3.8.1"

```

Uncaught RangeError: Maximum call stack size exceeded
    at JSON.stringify (<anonymous>)
    at resolveAndSanitizeRequestBodySchema (\dist\index.js:377)
    at \dist\index.js:371
    at Array.forEach (<anonymous>)
    at resolveAndSanitizeRequestBodySchema \dist\index.js:368)
    at \dist\index.js:371
```

jsdevel:
@AlexHolly not sure.  i haven't published your changes yet.

AlexHolly:
I don't think my changes have anything to do with it I use  "3.8.1" and this PR is in "3.8.2"

AlexHolly:
Even with > 4.0.0 this issue persists I am not sure what this code does but this solves it `if(!prop.$ref) {`.

```
else if ('properties' in requestBodySchema) {
        var schema_1 = requestBodySchema;
        Object.keys(schema_1.properties).forEach(function (property) {
            var prop = schema_1.properties[property];
            prop = sanitizeReadonlyPropertiesFromRequired(prop);
            if(!prop.$ref) {
                prop = resolveAndSanitizeRequestBodySchema(prop, v);
            }
        });
    }
```
It is trying to resolve $ref over and over again I am not able to run the tests, so If you think this is a possible solution, let me know I will open a PR.

jsdevel:
why can't you run the tests?

AlexHolly:
I got it working now, had to move the project to a different folder and use linux instead of windows ;)

Issue #455, opened 2019-05-20
example is missing in openapi-jsonschema-parameters


AlexHolly:
https://github.com/kogosoftwarellc/open-api/blob/master/packages/openapi-jsonschema-parameters/index.ts#L46
https://swagger.io/docs/specification/adding-examples/

jsdevel:
Please submit a PR

PR #478, opened 2019-06-22
More control over AJV

erayhanoglu:
AJV have some useful features like filtering and coercing. This patch allows us to use that useful options.

- [x] I only have 1 commit.
- [ ] I have added tests.
- [ ] My tests pass locally.

### DIFF
--- a/packages/openapi-response-validator/index.ts
+++ b/packages/openapi-response-validator/index.ts
@@ -22,6 +22,12 @@ export interface OpenAPIResponseValidatorArgs {
     [index: string]: IJsonSchema;
   };
   loggingKey: string;
+  useDefaults?: boolean | 'empty' | 'shared';
+  allErrors?: boolean;
+  unknownFormats?: boolean | string | string[];
+  removeAdditional?: boolean | string;
+  coerceTypes?: boolean | 'array';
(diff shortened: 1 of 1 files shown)


jsdevel:
not sure how i feel about this for a few reasons:

* this is an openapi framework, not a jsonschema framework per say.  ajv may have neat options, but what openapi use cases specifically are we addressing by exposing all the ajv options?
* ajv is the jsonschema validator we're using today, but we may want to go with a different validator in the future.  exposing ajv options couples us somewhat tightly to ajv
* ideally we should support the openapi spec directly without requiring the user to know how to configure ajv.

i'm leaning towards declining this pr, especially since the contributor guidelines weren't initially followed.  feel free to respond with your thoughts.

erayhanoglu:
Hi,
I know this is an open-api framework. But u know open-api just a standart for documenting APIs. Any framework can extend the standard. Both option is standard in GraphQL framework. Additionally graphql framework supports custom types, parsing and serializing. And this features saves developers days. This is why filtering and coercing is important.  This framework can help developers much more than this.

jsdevel:
if you find specific use cases for supporting the Open API spec, feel free to open a separate PR

Issue #368, opened 2019-02-25
Warning log : Ignoring the 2nd definition


fpozzobon:
Hi,

not sure if I am using the library correctly, we are trying to define the openApi using : 
- YAML file 
- Operations 

## Problem : 
we get the warning `express-openapi: /health.get has already been defined. Ignoring the 2nd definition...`

## Expected : 
no warning

## Setup : 
```javascript
return EOA.initialize({
      app,
      apiDoc: path.resolve(__dirname, 'openapi.yaml'),
      operations: {
        getHealth: (req, res) => {res.json({'result': 'ok'})}
      }
    });
```

openapi.yaml :
```
paths:
  /health:
    get:
      operationId: getHealth
      description: Health check
      responses:
        200:
          description: ''
```

## Question : 
Are we miss using the library ?

jsdevel:
@sebasmagri can you help on this one?

sebasmagri:
@jsdevel sure, I'm also getting those warnings... will fix those.

jsdevel:
thanks @sebasmagri !

Manny651:
I'm not sure if this should be considered as the same issue, but the same warning appears if we have an apiDoc with the same paths defined in args.paths.

Gino:
I also have the same issue, any updates on this?

sebasmagri:
@Manny651 you mean you have the same paths in the apiDoc with operations and in path directories?

I'm sorry this has taken so long, have been dealing with some other priorities on this end.

If anyone want to complete testing the PR linked above please feel free.

Manny651:
@sebasmagri yes, if we initialize with a complete apiDoc and use the "paths" options (not operations) to define the paths to the implementation, we get the same warning.

leedm777:
You can also see these warnings during the `basic-usage-with-central-apiDoc` test.

```
$ npx mocha packages/express-openapi/test/sample-projects/basic-usage-with-central-apiDoc/spec.js 
<snip/>
express-openapi: /users.delete has already been defined. Ignoring the 2nd definition...
express-openapi: /users.post has already been defined. Ignoring the 2nd definition...
express-openapi: /apiDocs.get has already been defined. Ignoring the 2nd definition...
express-openapi: /users/{id}.get has already been defined. Ignoring the 2nd definition...
express-openapi: /users/{id}.post has already been defined. Ignoring the 2nd definition...
<snip/>
```

jsdevel:
[...]

PR #479, opened 2019-06-22
openapi-framework: Suppressing unwanted "Ignoring the 2nd definition"…

seb-pereira:
ref: #368

This change suppress the unwanted _Ignoring the 2nd definition_ warning that is caused by a [duplicate reference of same method alias](https://github.com/kogosoftwarellc/open-api/commit/a19d5e9830698b569bd7f2e35cbdeaf04189c0b2#diff-acfbb4f9a52d04e45d5b249a6853c5ddR319) (one from module path definition and the other from the doc path definition).

The warning still makes sense when multiple aliases of the same method are defined. Example: `del`and `delete`, or `GET` and  `get`. I changed the logging statement to help correlate the duplicated element.

I have tested locally for non regression. No new test added as it only impact the logs. Note it does suppress the warning in test `basic-usage-with-central-apiDoc` (ref: [comment](https://github.com/kogosoftwarellc/open-api/issues/368#issuecomment-477845090))
 
- [x] I only have 1 commit.
- [ ] I have added tests **(see above)**
- [x] My tests pass locally.

### DIFF
--- a/packages/openapi-framework/index.ts
+++ b/packages/openapi-framework/index.ts
@@ -350,20 +350,23 @@ export default class OpenAPIFramework implements IOpenAPIFramework {
       this.apiDoc.paths[openapiPath] = pathDoc;
       const methodsProcessed = {};
 
-      [
-        ...Object.keys(pathModule).filter(byMethods),
-        ...Object.keys(pathDoc).filter(byMethods)
-      ].forEach(methodAlias => {
+      new Set(
+        Object.keys(pathModule)
+          .concat(Object.keys(pathDoc))
+          .filter(byMethods)
+      ).forEach(methodAlias => {
         const methodName = METHOD_ALIASES[methodAlias];
         if (methodName in methodsProcessed) {
           this.logger.warn(
             `${
               this.loggingPrefix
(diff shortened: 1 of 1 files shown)


jsdevel:
Thanks @seb-pereira!

petriaarnio:
"Example: del and delete": if api-docs.yml defines `delete`, how can I create an operation function for it when strict mode prevents using `delete` as function name but `del` or `DELETE` cause this warning?

jsdevel:
not sure @petriaarnio

agtb:
> "Example: del and delete": if api-docs.yml defines `delete` path, how can I create an operation function for it when strict mode prevents using `delete` as function name but `del` or `DELETE` cause this warning?

For anyone searching for the _delete has already been defined as ... Ignoring the 2nd definition_ warnings:

I'm doing it like this in the paths operations, exporting the delete function as a property (using ES6 modules):

```javascript
export default () => {
  const del = (req, res, next) => {
    // delete something
    return res.status(204).send()
  }

  const get = (req, res, next) => {
    // get something
    return res.status(200).json(something)
  }

  return {
    delete: del, // workaround for 'delete' is reserved word
    get,
  }
}
```

PR #481, opened 2019-06-26
openapi-response-validator: remove required writeOnly props from response validation

littldr:
Almost the same as #472 but now for props which are marked as required and `writeOnly`.

As defined [here](https://json-schema.org/latest/json-schema-validation.html#rfc.section.10.3) this allows to defined props which are required in the request, but must not be part of the response.

_Note_: For now there is no validation if the defined props is really not part of the response, it only removes them from the required array (in response validation). The same problem applies in the `request-validator` for `readOnly` props. 
Ajv [knows this keywords](https://github.com/epoberezkin/ajv#annotation-keywords), but does not validate them, do you have an idea how to add a proper validation for this?

### DIFF
--- a/packages/openapi-response-validator/index.ts
+++ b/packages/openapi-response-validator/index.ts
@@ -216,6 +216,16 @@ function recursiveTransformOpenAPIV3Definitions(object) {
     object.type = [object.type, 'null'];
     delete object.nullable;
   }
+  // Remove writeOnly properties from required array
+  if (object.properties && object.required) {
+    const writeOnlyProps = Object.keys(object.properties).filter(
+      key => object.properties[key].writeOnly
+    );
+    writeOnlyProps.forEach(value => {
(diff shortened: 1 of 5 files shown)


jsdevel:
> Ajv knows this keywords, but does not validate them, do you have an idea how to add a proper validation for this?

i don't

jsdevel:
@littldr this is a great PR!  thanks for following the Contributor Guidelines.  I'll merge once the build passes.  Right now it's failing on linting.

littldr:
> > Ajv knows this keywords, but does not validate them, do you have an idea how to add a proper validation for this?
> 
> i don't

Okay, I will try to figure a good way to do this and propose another PR.

I've fixed the lint error, so the build should now pass. :)

jsdevel:
Thank you so much @littldr!

PR #488, opened 2019-07-16
openapi-framework: Add support for $ref in requestBody (closes #486)

### LINKED ISSUE #486: v3: Consumes is not set correctly when operation refer to requestBody with a $ref
When setting the `requestBody` of an operation with a $ref such as:
```
// Operation
{
  requestBody: { $ref: '#/components/requestBodies/SomeBody' }
}

// Doc
{
  requestBodies: {
    SomeBody: {
       content: {
         'application/json': { schema: { type: 'object' }}
       }
    }
  }
}
```

The `consumes` property for the operation context fails to include the media type, making any `consumesMiddleware` for the media type silently not run for the operation.

This code seems to not resolve $refs if any before determining the consumed media types for operations.
https://github.com/kogosoftwarellc/open-api/blob/b01f1a637b2c621221a8b269f6047cd378b7c384/packages/openapi-framework/index.ts#L379-L388

jgranstrom:
This resolves any `$ref` in the operation's `requestBody` prior to determining what media types are consumed and before passing the definition to the validator so that `components.requestBodies` can be used in v3 documents when using validation and/or `consumesMiddleware`. Prior to this change `$ref` in `requestBody` would silently be treated incorrectly due to assumptions on it being defined inline.

### DIFF
--- a/packages/openapi-framework/index.ts
+++ b/packages/openapi-framework/index.ts
@@ -42,6 +42,7 @@ import {
   injectDependencies,
   METHOD_ALIASES,
   resolveParameterRefs,
+  resolveRequestBodyRefs,
   resolveResponseRefs,
   sortApiDocTags,
   sortOperationDocTags,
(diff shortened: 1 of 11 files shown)


jsdevel:
Awesome job @jgranstrom !  thanks!!!

PR #497, opened 2019-07-29
Fix OperationHandlerArray interface

Nox-404:
Fix typing error


jsdevel:
thanks!

jsdevel:
@Nox-404 can you check koa-openapi for the same issue?

PR #518, opened 2019-08-22
openapi-response-validator: validateResponse always returns an array (fixes #512)

### LINKED ISSUE #512: Same API for openapi-response-validator and openapi-request-validator 'validate'
I noticed a few differences, is it possible to make them more similar?

- Name both `validate` or `validateRequest / validateResponse`

  - **openapi-request-validator:**

    `validator.validate`

  - **openapi-response-validator:**

    `instance.validateResponse`

- The return type for `validate`:
   
  - in `request-validator` can be an Array or Object.

       https://github.com/kogosoftwarellc/open-api/blob/master/packages/openapi-request-validator/index.ts#L314

  - in `response-validator` is always an Object but sometimes the `errors` key is missing.
       https://github.com/kogosoftwarellc/open-api/blob/master/packages/openapi-response-validator/index.ts#L124
      https://github.com/kogosoftwarellc/open-api/blob/master/packages/openapi-response-validator/index.ts#L132

   It would be nice to always get an Array.

AlexHolly:
- [x ] My tests pass locally.

### DIFF
--- a/packages/openapi-response-validator/README.md
+++ b/packages/openapi-response-validator/README.md
@@ -15,7 +15,7 @@ See `./test/data-driven/*.js` for more examples.
 
 ```javascript
 var OpenAPIResponseValidator = require('openapi-response-validator');
-var instance = new OpenAPIResponseValidator({
+var responseValidator = new OpenAPIResponseValidator({
   responses: {
     200: {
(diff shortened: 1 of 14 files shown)


jsdevel:
i'll look further once tests are added

PR #519, opened 2019-08-22
openapi-request-validator: always return array (fixes #512)

### LINKED ISSUE #512: Same API for openapi-response-validator and openapi-request-validator 'validate'

AlexHolly:
- [x ] My tests pass locally.

### DIFF
--- a/packages/openapi-request-validator/README.md
+++ b/packages/openapi-request-validator/README.md
@@ -58,7 +58,7 @@ var request = {
   params: {},
   query: {foo: 'wow'}
 };
-var errors = validator.validate(request);
+var errors = requestValidator.validateRequest(request);
 console.log(errors); // => undefined
 ```
 
--- a/packages/openapi-request-validator/index.ts
+++ b/packages/openapi-request-validator/index.ts
@@ -206,9 +206,16 @@ export default class OpenAPIRequestValidator
       querySchema && v.compile(transformOpenAPIV3Definitions(querySchema));
   }
 
+  public validateRequest(request) {
+    return this.validate(request);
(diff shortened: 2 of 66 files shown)


jsdevel:
once tests are added i'll look again

AlexHolly:
It still doesn't look right, do you need the `status`?
https://github.com/kogosoftwarellc/open-api/pull/519/files#diff-32b5cc863d07f4a03bd815a69e54f78cR321
It should always have a status (200/400) or never, because an empty array means 200 ...

AlexHolly:
ping

jsdevel:
@AlexHolly i think the purpose behind the original object status was to support epress error middleware i.e. middleware with 4 parameters.

AlexHolly:
I could try to add a flag so it will keep the current api and only return an array if expected.

PR #528, opened 2019-09-22
openapi-request-validator: readme renamed validator variable (part of…

AlexHolly:
… #512)

### DIFF
--- a/packages/openapi-request-validator/index.ts
+++ b/packages/openapi-request-validator/index.ts
@@ -6,7 +6,7 @@ const contentTypeParser = require('content-type');
 const LOCAL_DEFINITION_REGEX = /^#\/([^\/]+)\/([^\/]+)$/;
 
 export interface IOpenAPIRequestValidator {
-  validate(request: OpenAPI.Request);
+  validateRequest(request: OpenAPI.Request);
 }
 
 export interface OpenAPIRequestValidatorArgs {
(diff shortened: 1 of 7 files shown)


packages/openapi-request-validator/index.ts
```diff
@@ -206,6 +206,13 @@ export default class OpenAPIRequestValidator
...
 
+  public validateRequest(request) {
+    return this.validate(request);
+  }
+
+  /**
+   * @deprecated use validateRequest instead.
+   */
   public validate(request) {
     const errors = [];
```
jsdevel: can you add a console.warn here and move the function body to validateRequest?

jsdevel: also need to update the tests to use validateRequest in all projects

AlexHolly:
It fails because the node_modules are not up to date, it should run fine after merging this PR.

jsdevel:
@AlexHolly i only merge passing builds

AlexHolly:
Do the tests use the local package?

@jsdevel 
> also need to update the tests to use validateRequest in all projects

This would only work after pushing the new version to npm.

Am I missing something? Should this be solved in two PR?

jsdevel:
> This would only work after pushing the new version to npm.

ahh, yes you're correct on that

AlexHolly:
looks like all cases have already been fixed.

PR #551, opened 2019-10-17
openapi-default-setter: Support default values from schema property

lenkan:
As described in the spec: https://swagger.io/docs/specification/describing-parameters/#query-parameters, parameter types are described using the `schema` property. The `default` property is defined under `parameter.schema.default`:

```
in: 'query',
name: 'someQuery',
schema: {
  type: 'string',
  default: 'someValue'
}
```

This PR implements support for this. But also keeps the support for reading the default value from the root level of the parameter object.


jsdevel:
wonderful.  thank you!

PR #605, opened 2019-12-16
koa-openapi: forward rejection from securityMiddleware (fixes #603)

### LINKED ISSUE #603: forward rejection from securityMiddleware
I'm using koa and would like to make use of securityHandlers. Please return the result of the handler in order to handle Promise rejections at
https://github.com/kogosoftwarellc/open-api/blob/4b2927ebf427471c54d42ff9648fb9f9d97095ef/packages/koa-openapi/index.ts#L210
```
return handler.handle(ctx, function (err, result) {
```

ameinhardt:


jsdevel:
@ameinhardt sorry for the late reply.  holidays here in the US.  can you add a test?

PR #615, opened 2020-01-21
openapi-request-validatior: readOnly properties should be rejected

kendzi:
ReadOnly properties should be present only in response. The request should reject them as those properties should not be modified.

JSON Schema Draft-07 brings new properties “readOnly”. That property allow to use conditional validation of object fields depending on context. That way the same schema definition could be re-use as part of post method where should be missing and get method where field is provided by server site. Here in openapi-request-validatior the context is always write, as we write data to server and all read only fields should be rejected.

### DIFF
--- a/packages/openapi-request-validator/index.ts
+++ b/packages/openapi-request-validator/index.ts
@@ -100,6 +100,28 @@ export default class OpenAPIRequestValidator
       logger: false
     });
 
+    v.removeKeyword('readOnly');
+    v.addKeyword('readOnly', {
+      modifying: true,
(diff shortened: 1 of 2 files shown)


jsdevel:
@kendzi can you remove package-lock.json from the commit?

kendzi:
Done

jsdevel:
thank you!

kendzi:
Would it be possible to make a release with that fix? Would be great if I could use it from the public repo.

jsdevel:
i'll try to get to it soon.

PR #636, opened 2020-03-05
openapi-jsonschema-parameters: Convert `examples` to an array (fixes #635)

### LINKED ISSUE #635: openapi-jsonschema-parameters: `examples`  should be an array
In all of the test cases that contain `examples`, the expected type of the output schemas of `examples` are `object`. However, in [JSON schema 7.0](https://json-schema.org/understanding-json-schema/reference/generic.html#annotations), `examples` is an array only. This problem makes the schema that uses `examples` fails the validation.

IvanaGyro:
In json schema, the field `examples` is an array, but in openapi, that field is an object. To correctly support `examples`, convert `examples` to an array.

### DIFF
--- a/packages/openapi-jsonschema-parameters/test/data-driven/convert-a-single-nullable-complex-query-param-to-json-schema.js
+++ b/packages/openapi-jsonschema-parameters/test/data-driven/convert-a-single-nullable-complex-query-param-to-json-schema.js
@@ -60,11 +60,7 @@ module.exports = {
               uniqueItems: false,
               enum: ['1', '3'],
               multipleOf: 57,
-              examples: {
-                example1: {
-                  value: 'asd'
-                }
-              }
+              examples: ['asd'],
             },
             {
               type: 'null'
--- a/packages/openapi-jsonschema-parameters/index.ts
+++ b/packages/openapi-jsonschema-parameters/index.ts
@@ -163,14 +163,18 @@ function getSchema(parameters, type) {
     schema = { properties: {} };
 
     params.forEach(param => {
+      let paramSchema;
       if ('schema' in param) {
-        const paramSchema = handleNullableSchema(param.schema);
-        if ('examples' in param && !('examples' in paramSchema)) {
-          paramSchema.examples = param.examples;
+        paramSchema = handleNullableSchema(param.schema);
+        if ('examples' in param) {
+          paramSchema.examples = getExamples(param.examples);
         }
         schema.properties[param.name] = paramSchema;
       } else {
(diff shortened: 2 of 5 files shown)


packages/openapi-jsonschema-parameters/test/data-driven/convert-a-single-nullable-complex-query-param-to-json-schema.js
```diff
@@ -60,11 +60,7 @@ module.exports = {
               uniqueItems: false,
               enum: ['1', '3'],
               multipleOf: 57,
-              examples: {
```
jsdevel: * i'm a bit concerned that we're breaking the format here.  is this used internally anywhere else in open-api?
* out of curiosity, what caused you to notice this?
* can you provide a link to examples in the openapi v3 spec?

IvanaGyro: Here are the links to examples in OpenAPI v3 spec.
- [In component object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#components-object)
- [In parameter obejct](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#fixed-fields-10)
- [In media type object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#media-type-object)

And here are the [example object itself](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#example-object) and the [JSON schema v7 spec](https://json-schema.org/understanding-json-schema/reference/generic.html#annotations).

Talking about the code you highlight, if I'm correct, `convert-a-single-nullable-complex-query-param-to-json-schema.js ` is the test data for OpenAPI v2 (swagger). I have read the specification of OpenAPI v2 and do not find any definition of `examples`, however, I'm not sure what will be the right strategy to convert the field mentioned in v3 but not mentioned in v2, so I make it stay in the original strategy.

This problem happened when I tried to add some examples in the parameter object in the main API spec which is named `apiDoc.yaml` in the official readme. The sample:
```yaml
openapi: 3.0.2
info:
  version: 1.0.0
  title: Title
components:
  parameters:
    sortQuery:
      in: query
      name: sort
      schema:
        type: string
      examples:
        single:
          value: key1
paths: {}
```

jsdevel: ok.  thanks @IvanaGyro!  i'm comfortable merging this.

PR #679, opened 2020-08-15
openapi-request-validator: fix: nullability for allOf, anyOf, oneOf

mastodon0:
```nullable: true``` validation didn't work on schemas like ```{
                nullable: true,
                allOf: [{ type: 'string' }]
              }``` because there was no type-property.

This change handles allOf, anyOf, oneOf. Please see the added tests that would fail to validate the requests on null-values.

### DIFF
--- a/packages/openapi-request-validator/index.ts
+++ b/packages/openapi-request-validator/index.ts
@@ -598,7 +598,7 @@ function sanitizeReadonlyPropertiesFromRequired(
 function recursiveTransformOpenAPIV3Definitions(object) {
   // Transformations //
   // OpenAPIV3 nullable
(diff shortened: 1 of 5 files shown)


jsdevel:
thank you!

npdev453:
@jsdevel why that fix are not publicated?

jsdevel:
@npdev453 it should be.  please open apr

npdev453:
@jsdevel, sorry, missed that there are two packages with simmilar names.

I repeated this changes for **openapi-response-validator** package:
https://github.com/kogosoftwarellc/open-api/pull/707

But in my mind perfectly it should be replaced by one shared library function call.

PR #680, opened 2020-08-17
Ignore TS build artifact

Marsup:
- [X] I only have 1 commit.
- [ ] I have added tests.
- [X] I'm using the latest code from the master branch and there are no conflicts on my branch.
- [X] My tests pass locally.
- [X] I have run linting against my code.

Recent versions of all the packages have been shipped with an artifact `tsc` produces, I'm assuming you didn't want to ship with that as it's useless to have it in the package. I'm also not very familiar with lerna so I hope this works.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -12,3 +12,4 @@ tsconfig.json
 *.swp
 *.spec.js
 .nyc_output
+tsconfig.tsbuildinfo


jsdevel:
good catch!  thanks!

PR #691, opened 2020-10-12
openapi-request-validator: Make `parameters` arg optional

dwrss:
For an openapi `OperationObject`, the `parameters` property is optional:
https://github.com/kogosoftwarellc/open-api/blob/983a1c79e54e9bcef6eea43fe22615591c919aa1/packages/openapi-types/index.ts#L97
It would be useful for the equivalent parameter of the validator to be optional so that a valid operation can be easily passed to the validator.

### DIFF
--- a/packages/openapi-request-validator/index.ts
+++ b/packages/openapi-request-validator/index.ts
@@ -18,7 +18,7 @@ export interface OpenAPIRequestValidatorArgs {
   };
   loggingKey?: string;
   logger?: Logger;
-  parameters: OpenAPI.Parameters;
+  parameters?: OpenAPI.Parameters;
   requestBody?: OpenAPIV3.RequestBodyObject;
   schemas?: IJsonSchema[];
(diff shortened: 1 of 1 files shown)


jsdevel:
can you add a test when it isn't given parameters?

dwrss:
> can you add a test when it isn't given parameters?

I believe [this one](https://github.com/kogosoftwarellc/open-api/blob/master/packages/openapi-request-validator/test/data-driven/accept-no-parameters.js) already covers it?

jsdevel:
@dwrss that has a request body though, won't that be considered a parameter?

dwrss:
> @dwrss that has a request body though, won't that be considered a parameter?

In OpenApi V3 `requestBody` and `parameters` are distinct keywords. They are also handled differently internally in this codebase.

This change is purely to make the Typescript definition match the implementation and the OpenApi spec.

I could write a test that passes no parameters and no request body but I'm not entirely sure that's useful?

PR #693, opened 2020-10-20
fs-routes: Also search for cjs files (#692)

### LINKED ISSUE #692: Cannot use express-openapi in ES Modules modus
If you use the package express-openapi in ES Modules (with import/export and `"type": "module"` in your package.json).
You get an error that you cannot use require, it is in the openapi-framework package.
I think this should be the fix.
See: https://nodejs.org/api/esm.html#esm_commonjs_json_and_native_modules

w3nl:
This code fixes #692
So if you use express-openapi or another package, you can run it in ES Modules modus.
If you dont's use cjs for the files, the require in openapi-framework doesnt work, because require must return a CommonJS module, and in ES Modules modus you should use the .cjs extension.
It is a very small fix, but very usefull.

If i change it to .cjs without this fix, the files are not found.

If i change the files to .js in ES Modules modus, you get this error:
```
(Use `node --trace-warnings ...` to show where the warning was created)
internal/modules/run_main.js:54
    internalBinding('errors').triggerUncaughtException(
                              ^

Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: .../api-routes/example.js
require() of ES modules is not supported.
require() of .../api-routes/example.js from node_modules/openapi-framework/dist/index.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename users.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from package.json.

    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1216:13)
    at Module.load (internal/modules/cjs/loader.js:1049:32)
    at Function.Module._load (internal/modules/cjs/loader.js:937:14)
    at Module.require (internal/modules/cjs/loader.js:1089:19)
    at require (internal/modules/cjs/helpers.js:73:18)
    at node_modules/openapi-framework/dist/index.js:141:37
    at Array.map (<anonymous>)
    at node_modules/openapi-framework/dist/index.js:137:26
    at Array.forEach (<anonymous>)
    at OpenAPIFramework.initialize (node_modules/openapi-framework/dist/index.js:122:19) {
  code: 'ERR_REQUIRE_ESM'
}
```

If you have a solution that it import all routes as ES Modules, that would be great, and better than this fix, because now all routes are still CommonJS.

### DIFF
--- a/packages/fs-routes/index.ts
+++ b/packages/fs-routes/index.ts
@@ -45,8 +45,8 @@ export default function fsRoutes(
   options: FsRoutesOptions = {}
 ): FsRoute[] {
   dir = path.resolve(process.cwd(), dir);
-  options.glob = options.glob || '**/*.js';
-  options.indexFileRegExp = options.indexFileRegExp || /(?:index)?\.js$/;
+  options.glob = options.glob || '**/*.+(js|cjs)';
+  options.indexFileRegExp = options.indexFileRegExp || /(?:index)?\.+(js|cjs)$/;
   const cacheKey = dir + options.glob;
 
   if (!memo[cacheKey]) {
--- a/packages/fs-routes/test/test.ts
+++ b/packages/fs-routes/test/test.ts
@@ -6,20 +6,40 @@ const testDir = path.resolve(__dirname, '..', 'test-dir');
 function assertRoutes(routes) {
   const output = [
     {
+      path: testDir + '/home.cjs',
+      route: '/home',
+    },
+    {
       path: testDir + '/home.js',
       route: '/home',
+    },
+    {
+      path: testDir + '/users/index.cjs',
+      route: '/users/',
     },
     {
       path: testDir + '/users/index.js',
       route: '/users/',
     },
     {
+      path: testDir + '/users/query.cjs',
+      route: '/users/query',
+    },
+    {
       path: testDir + '/users/query.js',
       route: '/users/query',
     },
     {
+      path: testDir + '/users/{id}.cjs',
+      route: '/users/{id}',
+    },
+    {
       path: testDir + '/users/{id}.js',
       route: '/users/{id}',
+    },
+    {
+      path: testDir + '/users/:id.cjs',
+      route: '/users/:id',
     },
     {
       path: testDir + '/users/:id.js',


packages/fs-routes/test/test.ts
```diff
@@ -5,22 +5,42 @@ const testDir = path.resolve(__dirname, '..', 'test-dir');
 
 function assertRoutes(routes) {
   const output = [
+    {
+      path: testDir + '/home.cjs',
+      route: '/home',
```
jsdevel, line 10: hmm, if there are 2 files next to each other with the same name but different extensions, does node use cjs over js?  if so, i would expect to not see the js routes here.

w3nl: Good feedback.
We can define what the expected result will be.
I should expect in a project you use 1 of both extensions, but we can filter out double matches.
I think if there are files with the same name, .js should be there default.

But if you have a better solution that we can use ES module routes, that will be also a lot better.
The require() in openapi-framework that dynamic load the files should use import().
It is for me to much refactoring or to less code knowledge, but if you can use import, that will be a better fix.

jsdevel: what does node do?  i don't feel you answered my previous question.

w3nl: .js is the default of you don't add an extension.
But I'm this package all files are found with glob, and than all files are used.

packages/fs-routes/index.ts
```diff
@@ -45,8 +45,8 @@ export default function fsRoutes(
   options: FsRoutesOptions = {}
 ): FsRoute[] {
   dir = path.resolve(process.cwd(), dir);
-  options.glob = options.glob || '**/*.js';
-  options.indexFileRegExp = options.indexFileRegExp || /(?:index)?\.js$/;
+  options.glob = options.glob || '**/*.+(js|cjs)';
```
jsdevel, line 48: i don't feel comfortable baking in specific extensions like this.  you can pass in options.glob for your specific use case.

ES modules are still experimental in node, and the semantics around the `cjs`, and `mjs` file extensions are a bit confusing to me: https://redfin.engineering/node-modules-at-war-why-commonjs-and-es-modules-cant-get-along-9617135eeca1

i'd say leverage `options.glob` for now.

w3nl: Ah, if we can set the glob with the express-openapi package, that would be better than my solution indeed.

PR #694, opened 2020-11-03
openapi-request-coercer: Implement coercion for deepObject style parameters

foxfriends:
OpenAPI 3.0 defines some serializations for objects/arrays in parameters. This implements support for the `deepObject` serialization by enabling coercers to be applied recursively based on the schema. 

The object serialization must still be enabled by the `enableObjectCoercion` flag, and if the style is not set to `deepObject`, then it uses the previous `JSON.parse` solution.


jsdevel:
thanks!

Issue #717, opened 2021-03-18
Bug with readOnly and required fields


artclark42:
Hello, I noticed a bug where if you have a required field and a read only field, the required field will not be validated against. 

As a sample 

```javascript
 schemas: {
      Test1: {
        type: 'object',
        properties: {
          foo: {
            type: 'string',
            readOnly: true,
          },
          bar: {
            type: 'string',
          },
        },
        required: ['bar'],
      },
    },

```
In this example, if I make a request without bar, the validation will pass without throwing an error. After doing some digging into the validator, it looks like in order to find the readOnly field in "required", the indexOf function is used with the read only property key. When this fails, it returns a -1. Since the readOnly field is not in the required list, the splice index gets set to -1. It looks like it doesn't check if this passes and just splices out whatever required field is at the -1th position, which is the last (and first in this case) property. 

I can issue a PR for this fix if that's okay. 

Thanks!

chadxz:
Hi @artclark42, welcome to Github! If you've already done all this work to determine the cause of the problem you are experiencing, it does make sense to submit a PR to fix it.

PR #719, opened 2021-03-18
openapi-request-validator: Added check for whether readOnly prop is i…

artclark42:
- Fixes issue #717 where required fields are ignored if there is a non-required readOnly property in the schema

### DIFF
--- a/packages/openapi-request-validator/index.ts
+++ b/packages/openapi-request-validator/index.ts
@@ -642,7 +642,9 @@ function sanitizeReadonlyPropertiesFromRequired(
       .filter((i) => i !== undefined)
       .forEach((value) => {
         const index = schema.required.indexOf(value);
-        schema.required.splice(index, 1);
+        if (index !== -1) {
+          schema.required.splice(index, 1);
+        }
       });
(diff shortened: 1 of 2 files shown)


packages/openapi-request-validator/index.ts
```diff
@@ -642,7 +642,9 @@ function sanitizeReadonlyPropertiesFromRequired(
       .filter((i) => i !== undefined)
       .forEach((value) => {
         const index = schema.required.indexOf(value);
-        schema.required.splice(index, 1);
+        if (index != -1) {
```
chadxz, line 645: looks like this is failing lint b/c it should be `!==` instead

chadxz (changes requested):
Fix the lint error and it looks good to me.

Does your test fail without your change?

chadxz (approved):

artclark42:
> Fix the lint error and it looks good to me.
> 
> Does your test fail without your change?

Yes the test fails without change. The linter script was not working for me locally otherwise I would have verified before opening the PR. Thanks!

PR #737, opened 2021-05-13
openapi-request-validator: add ajv options

nicofuccella:
This PR allow pass extra options to Ajv validator for example to enable removeAdditional.

Inspired by: https://github.com/kogosoftwarellc/open-api/pull/709

PR #738, opened 2021-05-13
[PENDING #737 MERGE] Express openapi/add request ajv options and response ajv options

### LINKED ISSUE #65: Option to strip out unused parameters
This may already be included somewhere, but it would be awesome if there was an option to strip out parameters that aren't in the docs. For example:

If a query parameter `age` is defined, but `?age=30&foo=bar` is passed, I would like `req.query` to only contain `age`.

I'm not sure which underlying `express-openapi-*` module would contain this, or perhaps it would be a new one.

nicofuccella:
Add `requestAjvOptions` and `responseAjvOptions` to `express-frameworks` constructor args.
This way you can easily set separate `ajv` params for request and response validations.
This is useful for example when you want to set `removeAdditional` for responses or requests (fixes #65):
```javascript
openapi.initialize({
  apiDoc: require('./api-doc.js'),
  app: app,
  paths: path.resolve(__dirname, 'api-routes'),
  requestAjvOptions: { removeAdditional: true }
});
```
Add two more tests in `express-openapi` to look at the usage.


giowe:
awesome! need this!

jsdevel:
please update with the latest from master.  ajv was bumped.

nicofuccella:
@jsdevel done

jsdevel:
@nicofuccella sorry for the delay here.  @asarver and i ran into issues getting the ajv bump published, but it should be good now with the latest.  can you update again please?  i'll merge as soon as you do so and the build is passing.  thanks!

jsdevel:
@nicofuccella can you update with the latest from master?  need to trigger a build with github actions

iggh:
I'm of the opinion https://github.com/kogosoftwarellc/open-api/pull/785 would be a cleaner way of providing your own defaults for request/response ajv options (and other features).

By being able to interrogate the args passed in on instantiation using your own class, it would be infinitely flexible to say, only use specific options for different base paths, or special case a single path that needs specific ajv options as part of the validation.

jsdevel:
@nicofuccella please consider what @iggh is suggesting and make a case for the approach outlined in this PR.

Issue #727, opened 2021-04-22
migrate to ajv >= 7 (ajv 6 has vulnerability and reaches end of life 30/06/21)


asarver:
The different open-api libraries are using v6 of AJV. There is an outstanding security vulnerability for version 6 (https://www.whitesourcesoftware.com/vulnerability-database/CVE-2020-15366) and while AJV says they'll fix security issues until June 30 2021, it's reaching end of life in 2 months.

It's not clear which JSON schema draft versions open-api supports (some places say draft 4, others don't specify a specific version), however AJV v6 supports draft 4, v7/v8 only support 6 and up. 

I can put up PRs for some of the changes required, but I'll need more information on how this impacts the different libraries/documentation/etc. 

main differences in new AJV version:
- the `.` separator for property names has changed to `/`
- array errors no longer have `[0]`, it's just `0`
- errors changed `should` to `must`
- `dataPath` changed to `instancePath`
- for custom keywords, `KeywordDefinition` has changed to use ` DataValidationCtx` that contains the parent schema data
- `id` is no longer supported (json schema draft 4) and instead needs to be `$id` (json schema draft 7)
- if using `format` in json schema, the library `ajv-formats` needs to be required now
- missing references always throws an error now where before it was an error open api showed through its normal errors
- positiveInteger has changed to nonNegativeInteger (which is great, the name is correct now)
- `unknownFormats` is no longer supported, and formats to ignore have to be explicitly ignored

asarver:
it looks like `swagger-schema-official` is using draft-04, but I can't find a repo anywhere to open up a PR. AJV can't be updated without moving off of draft-04

asarver:
since that schema hasn't been updated in 6 years, my current plan is to copy it into the openapi-schema-validator project and remove the swagger dep

asarver:
running into some issues with tests, but I'm hoping to have a PR opened tomorrow

asarver:
since this is a full library upgrade, it would make sense to be a breaking change version increment, or a major at the very least. I'll also have to create a PR for each package, since they are dependent on each other, and will need the latest version to pull in ajv v8

asarver:
just waiting for an open-api version bump, and I can close this issue out

asarver:
@jsdevel will those PRs be a part of the v9 release, or will a version bump be needed?

jsdevel:
[...]

PR #739, opened 2021-05-14
openapi-request-validator: upgrade to ajv version 8 for json schema 7 compatibility

asarver:
part of #727 (see checklist in issue of all changes from ajv6 -> ajv8)


asarver:
@jsdevel this and #740 should be the last two PRs to fully switch over to ajv v8

PR #747, opened 2021-06-07
openapi-types: allow array items to be reference objects

### LINKED ISSUE #746: [openapi-types] Property 'type' is missing in type '{ $ref: string; }' but required in type 'ItemsObject'.
The error above occurs, when I have an array schema, like this (example from the official swagger docs):

```
schema: {
  type: 'array',
  items: {
    $ref: '#/definitions/pet'
  }
}
```

The typings now expect items to include a type although it is a reference, which should not be the case.
This was not the case in version 8.0.0, but I couldn't find the change that caused this.
I believe the `SchemaObject` type should define `items?: ItemsObject | ReferenceObject;` instead of `items?: ItemsObject;`.

marikaner:
Fixes #746.

### DIFF
--- a/packages/openapi-types/index.ts
+++ b/packages/openapi-types/index.ts
@@ -568,7 +568,7 @@ export namespace OpenAPIV2 {
     externalDocs?: ExternalDocumentationObject;
     example?: any;
     default?: any;
-    items?: ItemsObject;
+    items?: ItemsObject | ReferenceObject;
     properties?: {
       [name: string]: SchemaObject;
     };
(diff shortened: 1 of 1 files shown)


jsdevel:
@marikaner sorry for the late reply.  please get the build to pass.

marikaner:
@jsdevel Also sorry for the late reply, please take a look and merge this PR if possible 😊

jsdevel:
@marikaner can you update this with the latest from master?  i had to migrate the build to github actions.

marikaner:
> @marikaner can you update this with the latest from master? i had to migrate the build to github actions.

Done!

Issue #618, opened 2020-01-26
[express-openapi] Additional Middlewares on routes


SkyaTura:
There is a way of adding additional middlewares on specific routes only?

I was able to add "global" middlewares by adding them to the root of my apiDoc, but it doesn't work if I add it to the method or the path I want.

saalaa:
Yes, there is an example of this in the documentation:

- https://github.com/kogosoftwarellc/open-api/tree/master/packages/express-openapi#typescript-example

This doesn't just work for Typescript of course. The whole idea is to provide an operation wrapped in a array along with its middleware(s):

export default function() {
    async function GET(req, res) {
        // ...
    }

    GET.apiDoc = {
        // ...
    };

    return {
        GET: [someMiddleware(), GET]
    }
};

SkyaTura:
But I use operationIds instead of the default auto-load, and passing arrays as and operation doesn't work

jsdevel:
@SkyaTura not sure what you mean by this:

> I use operationIds instead of the default auto-load

SkyaTura:
https://github.com/kogosoftwarellc/open-api/tree/master/packages/express-openapi#argsoperations @jsdevel

saalaa:
So, this shouldn't work any differently. A thorough reading of the [Express documentation](https://expressjs.com/fr/api.html#app.all) would make that clear for you.

> app.all(path, callback [, callback ...])
> [...]
>  Callback functions; can be:
>  -   A middleware function.
>  -   A series of middleware functions (separated by commas).
>  -   An array of middleware functions.
>  -   A combination of all of the above.

This code should do what you want:

```
initialize({
  app,
  apiDoc: './apiDoc.yml',
  operations: {
    getFoo: [
      someMiddleware(),
      function(req, res) {
        res.send('foo');
      }
    ]
  }
});
```

ealtunyay:
I will investigate further, but wanted to throw out there that this did not work for me. First, the type definition does not allow arrays for values of `operations`:
```
operations?: {
    [operationId: string]: (...arg: any[]) => any;
};
```

If I ignore it I get `TypeError: innerFunction.bind is not a function`:

```
initialize({
  app,
  apiDoc: path.join(__dirname, "/../../spec.v1.json"),
  operations: {
    // @ts-ignore
    postAuthLogin: [
      function(_req: any, res: any) {
        res.send("foo");
      }
    ]
  }
});
```

saalaa:
[...]

PR #769, opened 2021-09-23
openapi-framework: add middleware for operations

jjdonov:
The goal of this PR is to bring the functionality of `args.operations` up to speed with how `args.paths` works with regards to middleware.

While args.paths (autoload like functionality) allows you to expose an array of middleware as an operation handler, the args.operations requires that the operation handler be a function. Attempting to pass an array result in a `TypeError`

```
/path/to/project/node_modules/openapi-framework/dist/index.js:165
                                return innerFunction.bind({
                                                     ^

TypeError: innerFunction.bind is not a function
```

I believe this was first documented in #618, and can make mounting middleware awkward or impossible at times.  

This PR is may me incomplete. **If** the maintainers are open to a change like this, then I'd love to take it across the finish line. However, since I'm relatively new to the codebase, I expect I may need some guidance.

- [x] I only have 1 commit.
- [x] I have added tests.
- [x] My tests pass locally.

### DIFF
--- a/packages/openapi-framework/index.ts
+++ b/packages/openapi-framework/index.ts
@@ -258,14 +258,10 @@ export default class OpenAPIFramework implements IOpenAPIFramework {
               const operationId = methodDoc.operationId;
               if (operationId && operationId in this.operations) {
                 const operation = this.operations[operationId];
-
                 acc[METHOD_ALIASES[method]] = (() => {
                   const innerFunction: any = operation;
                   innerFunction.apiDoc = methodDoc;
-                  // Operations get dependencies injected in `this`
-                  return innerFunction.bind({
-                    dependencies: { ...this.dependencies },
-                  });
+                  return innerFunction;
                 })();
               } else if (operationId === undefined) {
                 this.logger.warn(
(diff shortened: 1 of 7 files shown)


packages/openapi-framework/index.ts
```diff
@@ -258,14 +258,10 @@ export default class OpenAPIFramework implements IOpenAPIFramework {
...
               if (operationId && operationId in this.operations) {
                 const operation = this.operations[operationId];
-
                 acc[METHOD_ALIASES[method]] = (() => {
                   const innerFunction: any = operation;
                   innerFunction.apiDoc = methodDoc;
-                  // Operations get dependencies injected in `this`
-                  return innerFunction.bind({
-                    dependencies: { ...this.dependencies },
-                  });
```
jjdonov, line 268: It _seems_ like this is  safe to remove, since [we inject dependencies later](https://github.com/kogosoftwarellc/open-api/blob/13ad704b645055d9baec6e047a2d1a1288edb538/packages/openapi-framework/index.ts#L323) on for all routes

jjdonov:
@jsdevel thanks for the merge 😄 hate to be a bother, but any idea when we might get a new version published? We have a lot of middleware work on the immediate horizon that this PR would help a ton with.

And in general, is there a roadmap/release calendar/similar available? We've become pretty heavy users of the library and would love to learn more about where its headed or ways to get involved!

jsdevel:
soon @jjdonov

Issue #778, opened 2021-11-24
Overwriting API doc for operation, when multiple operations share the same controller function


jjdonov:
#769 introduced a regression where multiple operations that share the same controller function now overwrite each others api doc, meaning that the last operation processed is on the only documentation that gets used.

While this feature was undocumented, we were making use of it. Since the framework does not guard against overwriting the api doc, I'm led to believe this was the intended behavior. This worked in previous version of the framework because of the `.bind` call 

I am working on a PR to fix, but could use some direction from maintainers

(from original PR)
> It seems like [`.bind`] is safe to remove, since we inject dependencies later on for all routes
>
> If its not safe, then
>
> - What test cases should I add to check this behavior?
> - How should bind be applied to an array of middlewares? Would we only bind the last element (the controller)?

I see two options:
1. Do the `.bind` and attach the apiDoc to each middleware in the array. This would change the `this` context for the middlewares which could be problematic in some cases.
2. Only do the `.bind` and attach apiDoc to the last middleware in the array, which most of the time should be the operation handler. If someone were to add an _after_ middleware, that may break the implementation. However, _after_ middlewares (outside of error handling middleware) isn't a common pattern in express from what I've seen. I am not sure about the other web frameworks that this library supports.

Thoughts on the above, or ideas for a cleaner solution?

---

Edit: [Here is a (failing) test that demonstrates the issue](https://github.com/jjdonov/open-api/commit/9548e67ad3eaef51a808e5e4fff097c780d7ef31)

PR #780, opened 2021-11-24
BugFix: #778 dont overwrite api doc when controller is shared

jjdonov:
This PR implements possible solution 1️⃣   from #778 which is a bug that was introduced in #769
It restores behavior of binding the operation, or (new behavior) by binding each middleware in the array.

I think it is worth some discussion about if this is the _correct_ fix and hopefully this can help kick start the conversation.

~All tests are currently passing, but linting is not. Just need to make some small tweaks to make prettier happy.~

### DIFF
--- a/packages/openapi-framework/index.ts
+++ b/packages/openapi-framework/index.ts
@@ -259,7 +259,19 @@ export default class OpenAPIFramework implements IOpenAPIFramework {
               if (operationId && operationId in this.operations) {
                 const operation = this.operations[operationId];
                 acc[METHOD_ALIASES[method]] = (() => {
-                  const innerFunction: any = operation;
+                  /**
+                   * We have two options:
+                   *
+                   * 1. the middleware gets bound + dependency injected, this may be breaking.
+                   * 2. we pick the last middleware as the operation handler. This means we cannot support
+                   *    _after_ middlewares (though not a common express pattern)
+                   */
+                  const innerFunction: any = Array.isArray(operation)
+                    ? operation.map((middleware) =>
+                        middleware.bind({ dependencies: this.dependencies })
+                      )
+                    : operation.bind({ dependencies: this.dependencies });
+
                   innerFunction.apiDoc = methodDoc;
                   return innerFunction;
                 })();
--- a/packages/openapi-framework/test/sample-projects/operations-dir-with-valid-methid-doc-after-dependency-injection/spec.ts
+++ b/packages/openapi-framework/test/sample-projects/operations-dir-with-valid-methid-doc-after-dependency-injection/spec.ts
@@ -18,6 +18,7 @@ describe(path.basename(__dirname), () => {
       name: 'some-framework',
       operations: {
         getFoo: require('./operations/foo'),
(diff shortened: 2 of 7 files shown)


packages/openapi-framework/index.ts
```diff
@@ -259,7 +259,19 @@ export default class OpenAPIFramework implements IOpenAPIFramework {
...
+                   * 2. we pick the last middleware as the operation handler. This means we cannot support
+                   *    _after_ middlewares (though not a common express pattern)
+                   */
+                  const innerFunction: any = Array.isArray(operation)
+                    ? operation.map((middleware) =>
+                        middleware.bind({ dependencies: this.dependencies })
+                      )
+                    : operation.bind({ dependencies: this.dependencies });
+
                   innerFunction.apiDoc = methodDoc;
```
jsdevel, line 275: i see, so you're just creating a copy of all the functions so that when apiDoc is assigned as a property it will be on the new copy not on the reference.  i think the implementation you have here is fine.  thank you!

packages/openapi-framework/test/sample-projects/operations-dir-with-valid-methid-doc-after-dependency-injection/spec.ts
```diff
@@ -45,6 +51,19 @@ describe(path.basename(__dirname), () => {
             },
           },
         });
+        expect(apiDoc.paths['/foo-two']).to.eql({
```
jsdevel, line 54: i think you should add a new controller that returns the operation doc, then verify that the 2 paths using the same controller return different docs according to the path they're mounted for.  ya know?

jsdevel (changes requested):
overall i'm good with the approach.  i left a comment about the test.  make that change and get the build to pass and i'll merge it.  thank you!

packages/openapi-framework/test/sample-projects/operations-dir-with-valid-methid-doc-after-dependency-injection/spec.ts
jjdonov: I feel like this test _mostly_ captures that already since both operations api docs get asserted.

I understand how to implement what you're asking for in express-openapi, but not sure how I'd do it here. I'll read around the codebase some more, but pointers would be appreciated.

jjdonov:
>  and get the build to pass and i'll merge it. thank you!

@jsdevel the failure is unrelated to these changes. I think any PR will fail the check. Its based on latest LTS and is now using node 16. It seems the new version of v8 that  node16 uses has a different error message for the same error 👎 

I can replicate this by switching my node version between 12 + 16 (didnt have 14 installed), and you can see the fix on 15d70b9

packages/openapi-framework/test/sample-projects/operations-dir-with-valid-methid-doc-after-dependency-injection/spec.ts
jjdonov: Implemented this in the express package for now.

jsdevel (approved):

PR #784, opened 2021-12-20
openapi-framework: Allow usage without implementation

### LINKED ISSUE #722: openapi framework throws uncaught error if an operation for api is not present
When an operation is defined in the OpenAPI spec, but doesn't have a matching function identified in the openapi-framework, the server crashes on startup because it tries to access an undefined property on the route item. Specifically here https://github.com/kogosoftwarellc/open-api/blob/master/packages/openapi-framework/index.ts#L358

image

It would be preferable to simply ignore the error and either just not process requests to that operation or return a 501 Not Implemented.

ncodernet:
This PR tried to resolve
#722 : It would be
nice to be able to use openapi-framework without an implementation, e.g.
to implement generic request validation.
Looking a the code it seem to have worked in the past, as
routeItem.operations rather looks like a typo.

### DIFF
--- a/packages/openapi-framework/index.ts
+++ b/packages/openapi-framework/index.ts
@@ -201,7 +201,7 @@ export default class OpenAPIFramework implements IOpenAPIFramework {
         : null;
 
     let paths = [];
-    let routes = [];
+    let routes: { path: string; module: any }[] = [];
     const routesCheckMap = {};
 
     if (this.paths) {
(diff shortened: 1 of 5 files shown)


jsdevel:
@ncodernet please look at the build log and ensure it passes

ncodernet:
@jsdevel : Sure, I have fixed the two lint issues. Could you please retrigger the voter build?

ncodernet:
Hello @jsdevel : Sorry for the delay and the failing voters. I'm working on Windows and therefore couldn't run the scripts . Now I have created a docker where I could run the script and now they are successfull for me. Could you please retrigger - I hope that they also work for you.

Issue #372, opened 2019-02-27
Initialize slow


MrKumaPants:
I'm loading a very large schema. It has about 500 paths, and about 1200 external references. I'm calling `initialize` like so:

```
openapi.initialize({
	apiDoc: this.apiDoc,
    app,
    enableObjectCoercion: true,
    externalSchemas: this.externalSchemas,
    paths: this.paths,
    promiseMode: true,
    validateApiDoc: true
});
```

`apiDoc` and `externalSchemas` are loaded from yaml files and sent to `initialize` as objects. One microservice uses about 34 paths. However it takes a consistent 26 seconds for initialize to finish.

Is there any way to speed this up on my end, in the library, etc?

jsdevel:
dang!  500 paths is a lot!

are the external references downloaded when your server starts up?  if so, i'd bet that's where most of your slowness is happening.

MrKumaPants:
Yeah, it is very large. We have microservices for different parts of the schema.

I do a `yaml.safeLoad(file)` for them all before sending them to `initialize.externalSchemas`. Reading the yaml files into JS objects only takes a few seconds.

MrKumaPants:
Edit: Ran a traffic monitor on my server. Looks like despite me supplying externalSchemas it's still downloading them from the web. I wonder if it's taking the full ref into account and not stripping the ref path?

jsdevel:
well, with 500 paths, i can only imagine how many operations you have.  if you were to `require` all those files without calling `.initialize`, how long does it take?

MrKumaPants:
This is all done loading yaml files and programmatically generating the path operations. Loading and generating all of that only takes 4 seconds or so.

jsdevel:
right, but during `.initialize`, it's going to call `require` for each of the paths right?

what's the output of this?

```
find ./paths-dir -type f | wc -l
```

jsdevel:
depending on how many require calls the framework has to make, it may make sense to use `args.operations`.

MrKumaPants:
I'm sending the paths to initialize as:

```
[
    {
        path,
        module: { get: Function() }
    }
]
```

jsdevel:
oh, well that changes everything lol

using `console`, can you log before and after times before and after the initialize call?

MrKumaPants:
Using this:

```
        console.log(new Date().toTimeString());
        openapi.initialize({
            apiDoc: this.apiDoc,
            app,
            enableObjectCoercion: true,
[...]

PR #785, opened 2022-01-23
openapi-framework: Allow injecting custom classes for features

jsau-:
The current implementation does not allow defining custom classes to be used for:
* defaults setting
* coercing
* request/response validation
* security handlers

This is absolutely fine for the general use case. 

However, in very large projects this incurs a fairly large penalty on startup. For example, to perform an integration test on a single endpoint all of the classes must be instantiated for all paths/operations.

Allowing consumers to inject their own instances of these classes will support behavior like, within test environments, providing thin wrappers around the existing classes for this functionality which are lazy-initialized on any property being accessed. Lazy-initializing these classes sees fairly large improvements in integration test times for large API documents.

This may be useful for other reports of slow startup for large API documents, eg. [Issue #372](https://github.com/kogosoftwarellc/open-api/issues/372).

This could be seen as a resolution to pull requests [#737](https://github.com/kogosoftwarellc/open-api/pull/737) and [#738](https://github.com/kogosoftwarellc/open-api/pull/737) as well. Instead of having to pass AJV configuration to `OpenAPIFrameworkArgs`, consumers are able to write their own classes for request/response validation that instantiate AJV as they see fit.

### DIFF
--- a/packages/openapi-framework/index.ts
+++ b/packages/openapi-framework/index.ts
@@ -75,6 +75,7 @@ export default class OpenAPIFramework implements IOpenAPIFramework {
   private enableObjectCoercion;
   private errorTransformer;
   private externalSchemas;
+  private features;
   private originalApiDoc;
   private operations;
   private paths;
@@ -156,6 +157,7 @@ export default class OpenAPIFramework implements IOpenAPIFramework {
     this.dependencies = args.dependencies;
     this.errorTransformer = args.errorTransformer;
     this.externalSchemas = args.externalSchemas;
+    this.features = args.features;
     this.operations = args.operations;
     this.paths = args.paths;
     this.pathsIgnore = args.pathsIgnore;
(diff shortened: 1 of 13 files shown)


honzapatCZ:
Bump

Any docs to set it up? Would like to use the removeAdditional: "all" option

jsau-:
@honzapatCZ I've not included any documentation on `express-openapi` or `koa-openapi` because the contrib guidelines say to keep things scoped to a single package; if this is approved I'll raise some extra ones to document in those packages after the fact.

In terms of how to use:

Let's say you're using `express-openapi`.

Currently you call the `initialize` function, eg.

```
initialize({
  app,
  apiDoc,
  ...
});
```

To use this you now have an extra parameter called `features`, where each property is a class which implements the same interface as the default package used by this library.

If you wanted to change the AJV behavior for request validation you would write a new class:

```
import {
  IOpenAPIRequestValidator,
  OpenAPIRequestValidatorArgs,
} from 'openapi-request-validator';

export default class CustomRequestValidator
  implements IOpenAPIRequestValidator {
  private args;

  constructor(args: OpenAPIRequestValidatorArgs) {
    this.args = args;
  }

  validateRequest(request: any) {
    return;
  }
}
```

(Except you'd probably copy the actual class code from `packages/openapi-request-validator/index.ts`, and make any changes desired.)

And then in initialize you'd change it to:

```
initialize({
  app,
  apiDoc,
  features: {
    requestValidator: CustomRequestValidator,
  },
  ...
});
```

jsdevel:
solid PR @jsau- !

PR #787, opened 2022-01-23
Express openapi/add request ajv options and response ajv options[Duplicate of #738]

honzapatCZ:
Duep of #738 to trigger the CI


iggh:
I'm of this opinion https://github.com/kogosoftwarellc/open-api/pull/785 would be a cleaner way of providing your own defaults for request ajv options (and other features).

By being able to interrogate the args passed in on instantiation using your own class, it would be infinitely flexible to say, only use specific options for different base paths, or special case a single path that needs specific ajv options as part of the validation.

honzapatCZ:
> I'm of the opinion #785 would be a cleaner way of providing your own defaults for request ajv options (and other features).
> 
> By being able to interrogate the args passed in on instantiation using your own class, it would be infinitely flexible to say, only use specific options for different base paths, or special case a single path that needs specific ajv options as part of the validation.

Completely agree. In fact I've created this before I stumbled upon that PR

PR #802, opened 2022-04-08
openapi-request-coercer: fix header coercing

jakobw:
Header field names are case insensitive which was addressed to some extent in the code by lowercasing the properties of the `coercers` object ([here](https://github.com/kogosoftwarellc/open-api/blob/5c68224a2e85bc1a55c7a6aa29c30e5d0f873032/packages/openapi-request-coercer/index.ts#L137) via the [`l` helper function](https://github.com/kogosoftwarellc/open-api/blob/5c68224a2e85bc1a55c7a6aa29c30e5d0f873032/packages/openapi-request-coercer/index.ts#L118)). However, the function performing the coercion compared the original (non-lowercase) fields to the lowercase coercers properties ([here](https://github.com/kogosoftwarellc/open-api/blob/5c68224a2e85bc1a55c7a6aa29c30e5d0f873032/packages/openapi-request-coercer/index.ts#L143)), which doesn't work unless the header was lowercase to begin with.

This is fixed here by using the same lowercasing on the parameter name when checking for a matching coercer.

### DIFF
--- a/packages/openapi-request-coercer/index.ts
+++ b/packages/openapi-request-coercer/index.ts
@@ -140,8 +140,8 @@ function buildCoercer(args) {
 
   return (obj) => {
     for (const paramName in obj) {
-      if (coercers.hasOwnProperty(paramName)) {
-        obj[paramName] = coercers[paramName](obj[paramName]);
+      if (coercers.hasOwnProperty(l(paramName))) {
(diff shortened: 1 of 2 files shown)


jakobw:
Thanks! :)

PR #809, opened 2022-05-17
openapi-request-validator: make mimetype header case insensitive

pulasthieka:
HTML headers are lowercase most of the time, but some applications (eg:AWS) may use a different case. In such cases headers like `content-type` can be sent as `Content-Type`. 

The request validator is written to throw an error when the request body is required AND  `content-type` header is not present. Currently this error is thrown if `content-type` header is not in lowercase. 

This fixes this issue and **makes the content-type header case-insensitive**

### DIFF
--- a/packages/openapi-request-validator/index.ts
+++ b/packages/openapi-request-validator/index.ts
@@ -297,7 +297,7 @@ export default class OpenAPIRequestValidator
     }
 
     if (this.requestBody) {
-      const contentType = request.headers['content-type'];
+      const contentType = getHeaderValue(request.headers, 'content-type');
       const mediaTypeMatch = getSchemaForMediaType(
         contentType,
         this.requestBody,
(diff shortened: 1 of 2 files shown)


packages/openapi-request-validator/index.ts
```diff
@@ -733,3 +736,14 @@ function transformOpenAPIV3Definitions(schema) {
   recursiveTransformOpenAPIV3Definitions(res);
   return res;
 }
+
+function getCaseInsensitiveHeaders(
+  requestHeaders: { [key: string]: any },
+  header: string
+) {
+  const pattern = new RegExp('(' + header + ')', 'i');
```
jsdevel, line 744: can you use `.toLowerCase` instead of `RegExp`?

jsdevel (changes requested):

jsdevel (changes requested):

packages/openapi-request-validator/index.ts
pulasthieka: That would restrict the code to only search for lowercase headers. But since used in that manner that seems better. Will make the change

pulasthieka:
Changes made

jsdevel:
@pulasthieka please get the build to pass

pulasthieka:
> @pulasthieka please get the build to pass

Changes made to code formatting and now passes local workflow.

pulasthieka:
@jsdevel can you trigger a PR build workflow

PR #810, opened 2022-05-19
Add support for ESM-based projects

robertkiel:
~~*Still WIP, postponed until release of Typescript 4.7*~~

Adds support for ESM-based projects 

Main change:

https://github.com/hoprnet/open-api/blob/4d00559ee5e445f1f176f494ebcc4a9e26da978c/packages/openapi-framework/index.ts#L233-L236

## Context

Within the Node.js ecosystem, there exist two package formats, *CommonJS* and the new de-facto standard *ECMAScript Modules* aka *ESM*.

CommonJS modues are usually loaded *synchronously* using `require()`, whereas ES modules are loaded either using the `import` keyword or are loaded dynamically, but *asynchronously*, using the builtin `import()` function.

To support loading both types of modules, the de-facto way to go is using `import()` as it supports both formats, CommonJS *and* ESM. Changing this turns various synchronous functions into asynchronous ones.

## Changes

- within `openapi-framework`, change from `require` to `import` of modules
- propagate new asynchronous API to all tests and dependent modules and their tests


jsdevel (approved):

jsdevel:
@robertkiel can you get the build to pass?

robertkiel:
> @robertkiel can you get the build to pass?

Yes, for sure. Currently awaiting the final release of Typescript 4.7 to have proper builds. Will add some commits in the next days.

robertkiel:
Managed to squeeze the last burning regarding default exports of CommonJS scripts and default exports of ECMAScript modules. Tests are supposed to run green now.

jsdevel (approved):

jsdevel:
thank you @robertkiel !! huge lift!

Issue #812, opened 2022-06-05
initialize() of express-openapi 12.0.0 always throws exception on Windows


renkei:
After the update to express-openapi 12.0.0 my CommonJS app crashes on Windows when I'm calling the `initialize` method from the `express-openapi` package:

```
Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file, data are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'
    at new NodeError (node:internal/errors:372:5)
    at throwIfUnsupportedURLScheme (node:internal/modules/esm/resolve:1120:11)
    at defaultResolve (node:internal/modules/esm/resolve:1200:3)
    at ESMLoader.resolve (node:internal/modules/esm/loader:580:30)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:294:18)
    at ESMLoader.import (node:internal/modules/esm/loader:380:22)
    at importModuleDynamically (node:internal/modules/cjs/loader:1043:29)
    at importModuleDynamicallyWrapper (node:internal/vm/module:437:21)
    at importModuleDynamically (node:vm:381:46)
    at importModuleDynamicallyCallback (node:internal/process/esm_loader:35:14)
```

For Windows, the issue can be solved by changing in [openapi-framework/index.ts, line 238](https://github.com/kogosoftwarellc/open-api/blob/master/packages/openapi-framework/index.ts#L238)

from
```typescript
const imported = await import(fsRoutesItem.path);
```
to 
```typescript
const imported = await import(`file:///${fsRoutesItem.path}`);
```
Of course, this is just a first dirty hack and not tested on other platforms than Windows. Can someone check this and provide a fixed version 12.0.1, please?

jsdevel:
@robertkiel can you help here?

robertkiel:
> @robertkiel can you help here?

Haven't tested the change on Windows yet ... The CommonJS script / ES module loader of Node.js seems to require some special treatment for Windows.

Will try on Windows tomorrow.

renkei:
Any update on this?

jsdevel:
@robertkiel would be great if you could provide some leadership/help here.

robertkiel:
> @robertkiel would be great if you could provide some leadership/help here.

Sorry for the late reply. Have been busy with a different issue - also related ESM migrations which brought some insights

[...]

PR #815, opened 2022-06-19
add absolute file prefix for dynamic imports

robertkiel:
Re https://github.com/kogosoftwarellc/open-api/issues/812

# Main change

Always add the `file://` prefix when loading ES moduls or CommonJS scripts.

Tested with:

- [x] Windows 11 (natively, no WSL)
- [x] Linux
- [x] MacOS , see https://github.com/kogosoftwarellc/open-api/runs/6957151421?check_suite_focus=true

# How to test

```sh
npm run bootstrap
cd packages/express-openapi
npm run test
```

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -10,6 +10,7 @@ jobs:
           node-version: lts/*
       - run: |
           npm ci
+          npm run bootstrap
           cd packages/express-openapi
           npm run test
 
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -2,16 +2,16 @@ name: PR Build
 on: pull_request
 jobs:
   code-quality:
-    runs-on: macos-latest
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
         with:
           node-version: lts/*
-      - run: |
-          npm ci
-          npm run bootstrap
-          cd packages/express-openapi
-          npm run test
+      - run: npm ci
+      - run: npm run lint
+      - run: npm run bootstrap
+      - run: npm run cover
+      - run: npm run combine-coverage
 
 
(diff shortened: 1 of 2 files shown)


7mllm7:
@robertkiel care you to explain the reason for this change?
This specific line basically fails my build (with webpack, when trying to bundle `openapi-framework` in a single bundle...).

I get:
```bash
ERROR in ./node_modules/openapi-framework/dist/index.js 132:47-84
Module not found: Error: Can't resolve 'file:/' in '/............./myproject/node_modules/openapi-framework/dist'
```

Which comes from exactly this line:
```js
const imported = await import(`file://${fsRoutesItem.path}`);
```

If I remove the `file://` bit, it's alright, but since you actually purposely added this, I don't want to change it.

Thanks!

robertkiel:
> @robertkiel care you to explain the reason for this change?
> This specific line basically fails my build (with webpack, when trying to bundle `openapi-framework` in a single bundle...).
> 
> I get:
> ```bash
> ERROR in ./node_modules/openapi-framework/dist/index.js 132:47-84
> Module not found: Error: Can't resolve 'file:/' in '/............./myproject/node_modules/openapi-framework/dist'
> ```
> 
> Which comes from exactly this line:
> ```js
> const imported = await import(`file://${fsRoutesItem.path}`);
> ```
> 
> If I remove the `file://` bit, it's alright, but since you actually purposely added this, I don't want to change it.
> 
> Thanks!

The Windows version of Node.js requires absolute paths when loading ES modules - in contrast to Linux-based OSes and MacOS which also accept a relative path.

At this point, it could be worth to have a special treatment for Windows-based machines.

Which machine and which version of Node.js are you running?

7mllm7:
@robertkiel thanks for your reply.
I'm running macOS Monterey (12.4) and node v14.19.2.

Any idea how I can avoid this error without forking 🤔?

Thanks!

robertkiel:
Probably upgrade to Node.js 16

coryasilva:
This should have been handled as @7mllm7 suggested.  

This change is breaking our jest tests.
````
Node 20
Mac local dev
Linux container
```

PR #817, opened 2022-06-24
extend expressJS Request Handler Parameters through express-openapi

joshAg:
The added generic types for RequestHandler come from expressJS and allow the user to give types to various parts of the request and response object in a way that allows typescript to infer types. Currently if the user wants typescript to infer one of these parts has a certain type they have to explicitly case the object as such. This will remove the need to cast at all. This change is necessary because with the current code the user cannot explicitly type req or res because the compiler will complain about assigning a function where they have been more defined to the Operation type since Operation is missing those specific type definitions.

For example

```
// normally generated from the openAPI definition
interface myExampleDictionary extends core.ParamsDictionary {
  key: string
}
const GET: Operation =  async (req: express.Request<myExampleDictionary>, res: express.Response, next: express.NextFunction) => {
```

Instead we currently have to do something like

```
// normally generated from the openAPI definition
interface myExampleDictionary extends core.ParamsDictionary {
  key: string
}
const GET: Operation =  async (req, res, next) => {
  const params = req.params as myExampleDictionary;
```

- [ ] I only have 1 commit.
- [ ] I have added tests.
- [ ] I'm using the latest code from the master branch and there are no conflicts on my branch.
- [ ] My tests pass locally.
- [ ] I have run linting against my code.

### DIFF
--- a/packages/express-openapi/index.ts
+++ b/packages/express-openapi/index.ts
@@ -1,3 +1,4 @@
+import * as core from 'express-serve-static-core';
 import { ErrorObject, FormatDefinition, Format } from 'ajv';
 import { Application, ErrorRequestHandler, RequestHandler } from 'express';
 import OpenAPIFramework, {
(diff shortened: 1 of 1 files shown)


jsdevel:
@joshAg please get the build to pass

PR #821, opened 2022-07-08
openapi-request-validator: Fix componentSchemas type (fixes #820)

### LINKED ISSUE #820: Incorrect type for componentSchemas in OpenAPIRequestValidator
Hello,

the `componentSchemas` field in `OpenAPIRequestValidator` is declared as:
```typescript
componentSchemas?: IJsonSchema[];
```
**But** looking to the tests, you are always passing an object instead of an array.
When I pass an object it works fine, but if I pass the schemas as array it fails to resolve references, with an error like:
```
Error: can't resolve reference #/components/schemas/TestItem from id #
```

Is it an idea to change the type to something like:
```Typescript
componentSchemas?: OpenAPIV3.ReferenceObject | OpenAPIV3.SchemaObject
```

aldex32:


jsdevel:
thanks @aldex32 .  i believe this is technically a breaking change correct?

aldex32:
Hey @jsdevel , yes it is a breaking change.
I can avoid it buy making it backward compatible like this:
```typescript
  componentSchemas?: IJsonSchema[] | { [key: string]: OpenAPIV3.ReferenceObject | OpenAPIV3.SchemaObject; };
```
 **But** I am not sure if the type `IJsonSchema[]` is/was correct. Also all the tests I see in this project are not assigning any array to this field. Unfortunately the tests are not type checking the arguments passed to the constructor of `OpenAPIRequestValidator` and this probably was the reason of not catching the wrong type.

jsdevel:
@aldex32 breaking change doesn't scare me.  is this in line with th OpenAPI schema?

jsdevel:
@aldex32 ?

PR #862, opened 2023-04-08
fs-routes: Lock glob version to 8.x.x for compatibility on Windows

rokoucha:
Since the combination of fs-routes and glob 9 or later stopped working on windows, I changed it to explicitly support up to glob 8.

<https://github.com/isaacs/node-glob/blob/main/changelog.md#90>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,7 @@
         "cors": "^2.7.1",
         "coveralls": "3.0.12",
         "express": "^4.13.3",
-        "glob": "7.1.6",
+        "glob": "8.1.0",
         "istanbul": "0.4.5",
         "istanbul-combine": "0.3.0",
         "lerna": "^3.22.1",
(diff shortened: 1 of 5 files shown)


Last-Order:
According to the docs, the backslash path delimiter is not supported by glob [since 8.x](https://github.com/isaacs/node-glob/blob/main/changelog.md#80:~:text=%5C%20is%20now%20only%20used%20as%20an%20escape%20character%2C%20and%20never%20as%20a%20path%20separator%20in%20glob%20patterns%2C%20so%20that%20Windows%20users%20have%20a%20way%20to%20match%20against%20filenames%20containing%20literal%20glob%20pattern%20characters.). And we can confirm that 8.x is [not working in projects using openapi-framework](https://github.com/Chinachu/Mirakurun/issues/132) anymore so maybe we should lock the version to 7.x?

rokoucha:
As far as I have tried, it works fine with glob 8.x.

> Paths are returned in the canonical formatting for the platform in question.
> https://github.com/isaacs/node-glob/blob/main/changelog.md#90
I think this change may have destroyed it.

jsdevel:
@rokoucha plz fix the build

rokoucha:
@jsdevel I fixed it. It seems that it was originally unstable, not broken by my change.

meyerds:
A better option to fix this issue might be to change `file` on [this line](https://github.com/kogosoftwarellc/open-api/blob/c805d4333c79891c29f8567dfc7f08da6334a64b/packages/fs-routes/index.ts#L58) to `path.posix.normalize(file)` which will avoid dependency on a specific glob version

jsdevel:
@rokoucha i agree with @meyerds .  locking down a dependency potentially opens up this project to dep lock in and security vulnerabilities in the long wrong.  can you try fixing it with the latest glob?

rokoucha:
OK, I will close this PR and try to fix it. thanks for the comments!

PR #865, opened 2023-05-02
express-openapi: add `TupleSchemaObject` (closes #864)

### LINKED ISSUE #864: Add support for `prefixItems`
From what I understand, OAS 3.1 should support the same validations offered by JSON Schema. It would therefore be possible to define a type `{ type: 'array', minItems: 2, maxItems: 2, prefixItems: [...] }` that would be used to define tuples.

[https://json-schema.org/draft/2020-12/json-schema-validation.html](https://json-schema.org/draft/2020-12/json-schema-validation.html)

I hope I'm not mistaken, but currently it is not supported in this repository:
https://github.com/kogosoftwarellc/open-api/blob/c805d4333c79891c29f8567dfc7f08da6334a64b/packages/openapi-types/index.ts#L148-L151

edobrb:
How i immagine the fix for #864.

### DIFF
--- a/packages/openapi-types/index.ts
+++ b/packages/openapi-types/index.ts
@@ -142,12 +142,18 @@ export namespace OpenAPIV3_1 {
    */
   export type SchemaObject =
     | ArraySchemaObject
+    | TupleSchemaObject
     | NonArraySchemaObject
     | MixedSchemaObject;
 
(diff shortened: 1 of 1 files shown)


jsdevel:
@edobrb can you write some tests to verify it works as you expect?

Issue #868, opened 2023-05-23
SyntaxError: Invalid regular expression during path-to-regexp conversion


ShadyAlexCodes:
Hi there,

I've encountered an issue when using express-openapi with my Express.js application. This application was based on a fresh installation.

The error I'm facing is as follows:
```
C:\Users\Alex\Visual Studio Code\TestProject\CustomerService\api\node_modules\express\node_modules\path-to-regexp\index.js:128
  return new RegExp(path, flags);
         ^

SyntaxError: Invalid regular expression: /^\/customer\(?:([^\/]+?))\/?$/: Unmatched ')'
```

This error occurs when my application is starting up.

Here is my project structure:
```
C:.
│   app.js
│   package-lock.json
│   package.json
│
├───api
│   │   api-doc.js
│   │
│   └───paths
│       │   test.js
│       │
│       └───customer
│               customer.js
│               {id}.js
│
└───bin
        www
```

Here is my package.json
```json
{
  "name": "customer-service",
  "version": "0.0.0",
  "private": true,
  "scripts": {
    "node": "nodemon app.js",
    "start": "node ./bin/www"
  },
  "dependencies": {
    "cookie-parser": "~1.4.4",
    "debug": "~4.3.4",
    "dotenv": "^16.0.3",
    "express": "~4.18.2",
    "express-openapi": "^12.1.1",
    "morgan": "~1.10.0",
    "nodemon": "^2.0.20",
    "swagger-ui-express": "^4.6.1"
  }
}
```

Here is my code for the `test.js`, `customer.js`, and `customer/{id}.js`
```js
module.exports = function () {
    let operations = {
        GET,
    };

    async function GET(req, res, next) {
        console.log("IN GET")
        console.log("REQUEST:", req)
        try {
            res.status(200).send();
        } catch (err) {
            console.error(err);
            res.status(500).json({
                error: "Internal Server Error"
            });
        }
    }

    GET.apiDoc = {
        summary: "Retrieve data from the api",
        operationId: "get-test",
        responses: {
            200: {
                description: "Returns if it was able to connect to the service",
            },
            500: {
                description: "Internal Server Error",
                content: {
                    "application/json": {
                        schema: {
                            $ref: "#/components/schemas/Error",
                        },
                    },
                },
            },
        },
    };
    return operations;
};
```

I am following the guidelines mentioned in the express-openapi documentation to structure my path files and use the {id} syntax for URL parameters.

[...]

PR #869, opened 2023-05-23
express-openapi: fixed windows path names

ShadyAlexCodes:
Windows path names are not being converted correctly. This modified function will take an OpenAPI path, replace all '{param}' style parameters with ':param' style (as used by Express), and also replace all backslashes (\) with forward slashes (/), making it compatible with all operating systems.

References the issue #868

### DIFF
--- a/packages/express-openapi/index.ts
+++ b/packages/express-openapi/index.ts
@@ -252,7 +252,8 @@ function optionallyAddQueryNormalizationMiddleware(
 }
 
 function toExpressParams(part) {
-  return part.replace(/\{([^}]+)}/g, ':$1');
+  // Replace '{param}' with ':param' and replace any '\\' with '/'
+  return part.replace(/\{([^}]+)}/g, ':$1').replace(/\\/g, '/');
 }
 
 function toPromiseCompatibleMiddleware(fn) {


ShadyAlexCodes:
I'm sorry for not getting back to you sooner. 

We can undoubtedly add tests for this change. It's worth noting that these tests don't necessarily need to be run on Windows. Since the changes we're making here are at the string transformation level, they are OS-agnostic and can be tested on any system.

However, I need to figure out how to incorporate these new tests into your existing testing framework. Therefore, I'd appreciate any guidance or suggestions you could give me.

renkei:
Instead of patching the npm package `express-openapi` I would suggest to patch the npm package `fs-routes`. What do you think?

For testing, I've modified node_modules/fs-routes/dist/index.js at line 44:

from
```
route: '/' + file.replace(options.indexFileRegExp, '')
```
to
```
route: '/' + file.replace(options.indexFileRegExp, '').replace(/\\/g, '/')
```
Then all other related openapi packages like openapi-framework, express-openapi, etc. don't need any patches. Swagger documentation looks also good again. At least for me, it works fine on Windows again, can someone confirm this?

mystiker:
Please drop this PR in favor of https://github.com/kogosoftwarellc/open-api/pull/877 and please merge it. Its a huge pain for us and easily fixed.

PR #876, opened 2023-05-26
openapi-response-validator: adds response arg to errorTransformer callback (closes #875)

### LINKED ISSUE #875: [ft. request] openapi-response-validator - allow access to response param in errorTransformer
### Context
* openapi-response-validator version: 12.1.3
* Node version:  18.12.1
* Package type: module (native node ESM ending in `.js`)

### Feature
Allow access to response param in errorTransformer so the error can be more informative. As a work-around I overwrite the `validateResponse` method to 'cache' the last response on `validator.lastTarget`, so I can log the erroneous value together with the rest of the error:
```js
const validator = new OpenAPIResponseValidator({
  components,
  responses: { '200': components.responses.Data },
  definitions: components.schemas,
  errorTransformer: function(openapiError, ajvError) {
    const keypath = openapiError.path.replace(/\//g, '.')
    const pathparts = keypath.split('.')
    let value = validator.lastTarget
      for (let i = 0; i < pathparts.length; i++) {
        if (!value) break
        value = value[pathparts[i]]
      }
    return `Error: Target ${keypath} ${openapiError.message}. Target ${keypath} value is: ${JSON.stringify(value, null, 2)}`
  }
})

// overwrite and store last validated target for error reporting
validator.validateResponse = function(statusCode, context) {
  validator.lastTarget = context
  return OpenAPIResponseValidator.default.prototype.validateResponse.call(this, statusCode, context)
}

console.log(validator.validateResponse(invalidData))
// {
//  message: 'The response was not valid.',
//  errors: [
//    'Error: Target data.items.2.instance must be object. Target data.items.2.instance value is: true'
//  ]
// }
```

webketje:
Also added some minor TS typing. Not sure what you mean by "linting" as there are no scripts/ linting configs defined in the repo.

I developed this on Windows and noticed all of the NPM scripts fail with `'..' is not recognized as an internal or external command,
operable program or batch file.` (even with Git Bash). I fixed this locally by running `export PATH=$PATH:"../../node_modules/.bin"` and replacing the package.json scripts (eg from `"test": "../../bin/mocha"` to `"mocha -r source-map-support/register -r ts-node/register test/**/*.ts"`).

I also noticed the unit tests use the dist/index.js.. That was kind of unexpected, as the test script loads `ts-node` anyway and is thus capable of loading `index.ts` directly (by doing `import Sut from '../index'` instead of `import Sut from '..'` in `test/data-driven.ts`).

### DIFF
--- a/packages/openapi-response-validator/index.ts
+++ b/packages/openapi-response-validator/index.ts
@@ -10,7 +10,7 @@ const LOCAL_DEFINITION_REGEX = /^#\/([^\/]+)\/([^\/]+)$/;
 
 export interface IOpenAPIResponseValidator {
   validateResponse(
-    statusCode: string,
+    statusCode: number|string,
     response: any
   ): void | OpenAPIResponseValidatorValidationError;
 }
@@ -35,7 +35,8 @@ export interface OpenAPIResponseValidatorArgs {
 
   errorTransformer?(
     openAPIResponseValidatorValidationError: OpenAPIResponseValidatorError,
-    ajvError: ErrorObject
+    ajvError: ErrorObject,
+    response: any
   ): any;
 }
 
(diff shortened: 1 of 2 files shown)


webketje:
Bump.

Issue #871, opened 2023-05-24
openapi-framework: fixed windows path names


rossi-p:
When using express-openapi to generate the Swagger file, the openapi-framework package is generating the paths with backslashes, as shown in the image below.

![image](https://github.com/kogosoftwarellc/open-api/assets/55264336/0c775aa4-3d9e-4912-a4a1-40fe805ce4f2)

Even though the PR #869 solves part of the problem, the docs are still generated with backslashes.

renkei:
I was able to fix this in the package `fs-routes` in file dist/index.js at line 44:

Change
```
route: '/' + file.replace(options.indexFileRegExp, '')
```
to
```
route: '/' + file.replace(options.indexFileRegExp, '').replace(/\\/g, '/')
```
Then the entire openapi-framework, express-openapi, etc. packages don't see Windows backslashes anymore. Swagger looks good again.

renkei:
I've created a [PR #877](https://github.com/kogosoftwarellc/open-api/pull/877)

renkei:
I think, I've found a better solution. Instead of calling `replace(/\\/, '/')` to get rid of backslashes for routes on Windows, we can use the glob option `posix: true` that was introduced with glob 10.1.0, this I've found in the [glob changelog](https://github.com/isaacs/node-glob/blob/main/changelog.md)

> Add posix: true option to return / delimited paths, even on Windows.

 I've already updated my PR where I've also updated package.json and package-lock.json to request the latest glob version >= 10.2.7

PR #877, opened 2023-06-02
fs-routes: Fixes backslash issue on Windows

### LINKED ISSUE #868: SyntaxError: Invalid regular expression during path-to-regexp conversion

renkei:
Fixes #868 and #871

### DIFF
--- a/packages/fs-routes/index.ts
+++ b/packages/fs-routes/index.ts
@@ -55,7 +55,7 @@ export default function fsRoutes(
       .sort(compare)
       .map((file) => ({
         path: path.resolve(dir, file),
-        route: '/' + file.replace(options.indexFileRegExp, ''),
+        route: '/' + file.replace(options.indexFileRegExp, '').replace(/\\/g, '/'),
       }));
   }
 


renkei:
I've found out that downgrading glob from version 10 to 7 also fixes the issue. Currently, package `express-openapi` comes with a `"glob": "*"` dependency, changing this into `"glob": "7"` would also solve the issue. Anyway, I still prefer this PR to be independent of the format returned by glob.

renkei:
I've update my PR because I've found probably a more elegant solution. Instead of calling `replace(/\\/, '/')` to get rid of backslashes on Windows for routes, we can use the glob option `posix: true` that was introduced with glob 10.1.0, this I've found in the [glob changelog](https://github.com/isaacs/node-glob/blob/main/changelog.md)

> Add posix: true option to return / delimited paths, even on Windows.

I've also updated package.json and package-lock.json to request the latest glob version >= 10.2.7

alex77h:
would love to see this merged. hope this project isnt abandoned - long time no update...

anthonyloukinas:
Would also love this merged. I tested this on Windows and it fixes the issues I was having using `{id}.js` file formatting (Using the curly braces).

```
.sync(options.glob, { cwd: dir, posix: true })
```

Setting that `posix: true` works.

mystiker:
Please merge this!

mystiker:
> I've found out that downgrading glob from version 10 to 7 also fixes the issue. Currently, package `express-openapi` comes with a `"glob": "*"` dependency, changing this into `"glob": "7"` would also solve the issue. Anyway, I still prefer this PR to be independent of the format returned by glob.

Can you please elaborate how you done this? I tried to set the following in my package.json in a project referencing `express-openapi`:

```
  "overrides": {
    "express-openapi": {
      "glob": "7"
    }
  },
```

But to no avail. Might be because I am using a somewhat complex monorepo with npm modules and maybe the resolution does not work as I expect.

PR #899, opened 2024-03-07
[WIP] feat: support x- specification extensions everywhere without a pre-existing index signature

mitchell-merry:
Per the [Swagger 2.0](https://swagger.io/specification/v2/), [OpenAPI 3.0](https://swagger.io/specification/v3/#specification-extensions), and [OpenAPI 3.1](https://swagger.io/specification/#specification-extensions) specs, most objects should support `^x-` properties for the purposes of extending the specification. However, `openapi-types` does not support these yet.

This PR introduces the `Extensible` interface, which is used to specify the types that should support this. It also bumps the prettier package to 2.2.0, which is the earliest version which supports TypeScript templating strings.

As is, this PR contains breaking changes to the types, to better fit the specfication! See details in the comments


packages/openapi-types/index.ts
```diff
@@ -92,9 +97,10 @@ export namespace OpenAPIV3_1 {
   >;
 
   export type PathsObject<T extends {} = {}, P extends {} = {}> = Record<
-    string,
+    `/${string}`,
```
mitchell-merry, line 100: ⚠️ This is a breaking change! We constrain the key to be `/${string}`, as the specification states.

We need to do this because otherwise this index signature (generic `string`) would shadow the `x-${string}` signature.

packages/openapi-types/index.ts
```diff
@@ -234,6 +243,7 @@ export namespace OpenAPIV3_1 {
     }
   >;
 
+  // TODO comment
   export type CallbackObject = Record<string, PathItemObject | ReferenceObject>;
```
mitchell-merry, line 247: This is extensible according to the specification, but a limitation of TypeScript means we can't add the `x-${string}` signature as it would be shadowed by the `string` one (which we can't constrain any further).

packages/openapi-types/index.ts
```diff
@@ -509,6 +523,7 @@ export namespace OpenAPIV3 {
     server?: ServerObject;
   }
 
+  // TODO comment
   export interface CallbackObject {
```
mitchell-merry, line 527: See https://github.com/kogosoftwarellc/open-api/pull/899/files#r1516178915

packages/openapi-types/index.ts
```diff
@@ -641,6 +656,7 @@ export namespace OpenAPIV2 {
     | SecuritySchemeOauth2Password
     | SecuritySchemeOauth2Application;
 
+  // TODO comment
   export interface ScopesObject {
```
mitchell-merry, line 660: See https://github.com/kogosoftwarellc/open-api/pull/899/files#r1516178915

packages/openapi-types/index.ts
```diff
@@ -817,12 +829,13 @@ export namespace OpenAPIV2 {
   }
 
   export interface ExternalDocumentationObject {
+    // TODO comment
     [index: string]: any;
```
mitchell-merry, line 833: See https://github.com/kogosoftwarellc/open-api/pull/899/files#r1516178915

packages/openapi-types/index.ts
```diff
@@ -843,6 +856,7 @@ export namespace OpenAPIV2 {
     $ref?: string;
   }
 
+  // TODO comment
   export interface XMLObject {
```
mitchell-merry, line 860: See https://github.com/kogosoftwarellc/open-api/pull/899/files#r1516178915

packages/openapi-types/index.ts
```diff
@@ -182,7 +188,8 @@ export namespace OpenAPIV3_1 {
     }
   >;
 
-  export type DiscriminatorObject = OpenAPIV3.DiscriminatorObject;
+  export type DiscriminatorObject = OpenAPIV3.DiscriminatorObject &
```
mitchell-merry, line 191: This is extensible in 3.1, but not in 3.0 (or at least the specification doesn't explicitly call this out).

jsdevel:
@mitchell-merry can you resolve conflicts and ensure a passing build?

mitchell-merry:
Merge conflict resolved, though it looks like you need to approve the build to be run

PR #908, opened 2024-09-10
openapi-framework: Fix dynamic import to work with jest (fixes #909)

### LINKED ISSUE #909: openapi-framework: dynamic import breaks jest transpile
When using jest this code gets transpiled to require and the file URL syntax of the import is not supported and breaks the testing environment. This change only uses file URL import syntax on windows machines. This fixes our tests and is still supported on linux and mac operating systems.

Why not just use jest's --experimental-vm-modules flag? Glad you asked. Because of this https://github.com/jestjs/jest/issues/11434#issuecomment-1410445848 which is caused by this https://github.com/nodejs/node/issues/37648 which is ultimately caused by this [v8 bug](https://issues.chromium.org/issues/40784051)

coryasilva:
When using jest this code gets transpiled to `require` and the file URL syntax of the import is not supported and breaks the testing environment.  This change only uses file URL import syntax on windows machines.  This fixes our tests and is still supported on linux and mac operating systems.

I think windows is definitely in the minority here and we should not optimize for it.  Furthermore, I believe using a UNC path would be a better solution outright but I cannot test because I don't have a window machine at hand.

Why not just use jest's `--experimental-vm-modules` flag? Glad you asked.  Because of this [jest bug](https://github.com/jestjs/jest/issues/11434#issuecomment-1410445848) which is caused by this [node bug](https://github.com/nodejs/node/issues/37648) which is ultimately caused by this [v8 bug](https://issues.chromium.org/issues/40784051)

- [x] I only have 1 commit.
- [ ] I have added tests. (No, should be covered by existing regression tests)
- [x] My tests pass locally.

### DIFF
--- a/packages/openapi-framework/index.ts
+++ b/packages/openapi-framework/index.ts
@@ -1,3 +1,4 @@
+import os from 'node:os';
 import fsRoutes from 'fs-routes';
 import OpenAPIDefaultSetter from 'openapi-default-setter';
 import OpenAPIRequestCoercer from 'openapi-request-coercer';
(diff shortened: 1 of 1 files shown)


amp5208:
confirmed, this is an issue currently that arises when trying to integrate [express-openapi](https://www.npmjs.com/package/express-openapi) into [backstage](https://github.com/backstage/backstage)

PR #920, opened 2026-04-21
[codex] Add OpenAPI 3.2 type definitions

closedLoop:
## Summary

- add a new `OpenAPIV3_2` namespace alongside the existing OpenAPI 3.0 and 3.1 definitions
- include OpenAPI 3.2 additions such as `query`, `additionalOperations`, `itemSchema`, `oauth2MetadataUrl`, `deviceAuthorization`, response summaries, and richer tag metadata
- widen the top-level `OpenAPI.Document`, `Operation`, and parameter unions so consumers can work with 3.2 documents directly

## Why

The package currently stops at OpenAPI 3.1, which forces downstream libraries to augment these types locally before they can adopt OpenAPI 3.2. This change brings the published type surface forward so consumers can target 3.2 directly.

## Validation

- `npm run prepare` in `packages/openapi-types`
- `./node_modules/.bin/tslint -c tslint.json packages/openapi-types/index.ts`


closedLoop:
@jsdevel could you please take a look?

PR #921, opened 2026-06-24
Bump js-yaml to address vulnerabilities in openapi-framework

jorgedelamora:
## Issue: 
package js-yaml vulnerability in versions lower than 4.2.0
https://github.com/advisories/GHSA-h67p-54hq-rp68

## Description
I need to use a version of openapi-framework that uses js-yaml 4.2.0 or higher, in order to override this package and address vulnerabilities. The problem is that openapi-framework uses the method `jsYaml.safeLoad` which is deprecated in v4.

This Pr bumps js-yaml to version 4.2.0 and changes the deprecated `jsYaml.safeLoad` to the safer `jsYaml.Load` method

### before
Screenshot 2026-06-24 at 8 10 05

### after
Screenshot 2026-06-24 at 8 12 08

### DIFF
--- a/packages/openapi-framework/package-lock.json
+++ b/packages/openapi-framework/package-lock.json
@@ -1,38 +1,39 @@
 {
   "name": "openapi-framework",
   "version": "12.1.3",
-  "lockfileVersion": 2,
+  "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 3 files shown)


blockwooddev:
I'm encountering this issue as well. It would be very helpful to have this merged and tagged!

